PR #290, opened 2014-12-06
Super Fancy Map Markers

lislis:
Sarah and I tried to make map markers on the command line. And that worked!
Now we need to do that in carrierware/ minimagick dsl. HELP!


PragTob:
This is a bit old... but any chance we can get it back up to speed again? :D

lislis:
Plan was actually to ditch this PR and then start fresh with the dragonfly gem. This is more or less or reference

PragTob:
If it is more or less a reference I'll close it, it can still be found, the branch is ther,e but just closed so we are not confused :)

lislis:
that's cool. This needs to be redone completely anyway

PR #299, opened 2015-01-16
Omniauth

FrauBienenstich:
this branch still needs some more love i think


berlintam:
eeeeermageeeeerd

berlintam:
small update: I created an application on github and have all da keyz.
we tried to set up a secrets.yml on shelly cloud, but couldn't get it to work
https://shellycloud.com/documentation/configuration_files

I think that getting the keys up on shelly cloud is the most important thing to do before this is merged. (AND TESTS SAYS SUSANNE. DO THEM THEN. GAWD)

berlintam:
ok so.... we merged this but then we in the process broke everything... so we rolled it back. Problem is possibly the syntax of the keys on the server in the secrets.yml. we had
```
production:
   github_secret: sdfsdfsdf
   github_id: klkjlkjlkj
```

It might also be `Unexpected error while processing request: Missing `secret_key_base` for 'development' environment, set this value in 'config/secrets.yml'`. We got this locally, and it seems as though it also probably happened on production. 

We don't have a config/secrets for our app because it was introduced in Rails 4.1 or maybe 4.2... there are conflicting stories out there on the internetz. We are thinking we should make one for our app....We also think we should upgrade rails. 

We are also confused and put everything back the way it was.

til:
secrets.yml was added in 4.1, see
http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#config-secrets-yml
for the non-conflicting truth.

The "missing secret_key_base" error is weird - I don't have that value
set in the app with secrets.yml, and it works fine without. It's set
instead in config/initializers/secret_token.rb, just like in the
rorganize.it app.

Upgrading to rails 4.2 is certainly a good thing to do one day, but I
don't see that this branch requires it.

FrauBienenstich:
ok, i will try this AGAIN!

PR #304, opened 2015-02-07
SR delete group 301

sareg0:


til:
This branch was accidentially branched off from an older version of the markers branch, not master. Therefore this PR was deleted.

Issue #301, opened 2015-02-03
delete a group when it no longer exists


lislis:
admins can delete groups, but if a group decides to no meet no longer, there should be a way

sareg0:
Hey guys,
Just to protect against malicious people that may have fallen out with their group and want to delete it themselves, is the 'delete group' function for non-members on a 'request' basis, or will it be instantly deleted?

Just wanted to make sure I understand what functionality we want on this :)

til:
Apparently deleting a group from the admin dashboard does not work. @sareg0 is on it

lislis:
Since this seems to be merged I'll close this

til:
Yes, I think this is fixed. Thanks @lislis!

PR #305, opened 2015-02-07
fixed broken admin delete group function #301

sareg0:


til:
Hooray!

sareg0:
As allowing normal users permissions to delete the group could be problematic, we wanted to make the request for deletion happen via email. Perhaps this process could be improved at a later date, BUT for now we thought it would not only an easy win, but also 'safer'


We recalled this email from Tam, and wanted set up an admin email.

**"So, yesterday when I got you to type your passwords into my terminal, I
was setting up your email addresses for you!

You all now each have an email address which is:
your_name_here@rubycorn.sabic.uberspace.de

If you use an email client such as thunderbird or apple mail, you can
configure it to also include this email address which you can use to
send and receive emails. If any of you are coming to the cryptoparty
tonite, you will soon have thunderbird, and I can show you how to do this.

Soon thou, your email addresses will be your_name_here@rubycorns.club.
HURRAY!"**

@lislis or @berlintam could you please set up the email roar@rorganize.it or something?

We were hoping that the 'roar' will be the mascot bot that will be our little spokesperson via email and such. 

But of course you can use your judgement. 

We are going to set up the email link now, but will wait to go live with the changes, until the email address has been setup. Will just us a 'testing' address for now.

berlintam:
awesome!

Maybe we can do this on tuesday? The rubycorns site still isn't working and I need help figuring that out, and it fits in with this topic (the topic being lisa-please-explain-and-help-with-gandi-and-uberspace)

til:
sareg0 <notifications@github.com> writes:
> @lislis or @berlintam could you please set up the email
> roar@rorganize.it or something?

The app is hosted on Shelly Cloud and the domain rorganize.it points
there. The Shelly Cloud server cannot receive incoming emails, so it is
my understanding that we need to set an DNS record of type MX that
points to a different server, which can receive incoming email. To have
one domain point to different hosts for web and for email is perfectly
possible and normal. I suggest to use the uberspace account for that, so
the DNS record should probably look like this:

rorganize.it.    MX    sabic.uberspace.de.

Priority doesn't matter because it will be the only MX entry.

And then follow parts of the instructions here

https://wiki.uberspace.de/domain:verwalten

and read this:

https://wiki.uberspace.de/mail:dotqmail#nutzung_mit_eigener_domain

On uberspace we can then freely configure to relay incoming emails
addressed to roar@rorganize.it to one or more persons. Maybe even to the
tuesdays group mailing list, but I am not sure it will accept them
without the sender being subscribed.

PR #307, opened 2015-02-07
Remove person from group, closes #300

### LINKED ISSUE #300: remove people from group
It might happen that people join groups but never show up, or show up and don't go to the group anymore. Members of that group might want to remove those people from the group.

lislis:
this adds links for admins in the group#show and person#show views to remove the person from the group

### DIFF
--- a/app/views/groups/_remove_from_group.html.haml
+++ b/app/views/groups/_remove_from_group.html.haml
@@ -1,0 +1,6 @@
+= form_tag membership_path(object.memberships.find_by(group_id: group.id)),
+  method: :delete, 
+  class: 'inline pull-right', 
+  data: { confirm: "This will remove the person from the group.\nYou really want to do that?" } do
+  %button.remove-from-group.no-btn.link
+    = fa_icon("times")
--- a/spec/features/admin_remove_person_spec.rb
+++ b/spec/features/admin_remove_person_spec.rb
@@ -1,0 +1,51 @@
+require 'spec_helper'
+
+feature 'Remove a person from a group' do
+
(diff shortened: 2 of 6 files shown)


app/views/groups/_remove_from_group.html.haml
```diff
@@ -0,0 +1,6 @@
+= form_tag membership_path(object.memberships.find_by(group_id: group.id)),
+  method: :delete, 
+  class: 'inline pull-right', 
+  data: { confirm: "This will remove the person from the group.\nYou really want to do that?" } do
+  %button.remove-from-group.no-btn.like-link
+    =fa_icon("times")
```
til: Haml wants a space after the `=` ... and everyone wants a newline at the end of each file.

spec/features/admin_remove_person_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      expect(page).to_not have_selector('.remove-from-group')
+    end
+
+  end
+end
+
+def go_to_group_page
+  visit '/groups/'
+  click_link 'Test Group'
+end
```
til: I would put this method definition inside the `feature` block, otherwise it will be defined globally.

til:
Alternatively, `link_to ... :method => :delete` could be used: http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

til:
Ok, `link_to` would work, but is the same effort as `form_for` and not as nicely gracefully degrading. `form_for` also works without JavaScript, so leaving it like this.

PR #309, opened 2015-02-07
add statistics, closes #286

### LINKED ISSUE #286: [CTA] Statistics for the start page
Good afternoon eveyone,

I'd like to suggest a counter for some statistics at the start page similiar to this:

<START>

Already registered in our Rails Girls App:

<strong>70 students | 20 coaches | 10 project groups</strong>

Sign up! (Button)
<END>

Example: http://screencast.com/t/9OTYEoDMx4

What do you think?

Best,
Thuy

lislis:
Some counter stuff for the home page and a little of design-ish stuff

### DIFF
--- a/app/assets/stylesheets/custom/_bootstrap_vars.scss
+++ b/app/assets/stylesheets/custom/_bootstrap_vars.scss
@@ -13,5 +13,4 @@ $btn-primary-color: white;
 $btn-primary-bg: $header_red;
 $btn-info-bg: $header_red;
 
-$jumbotron-bg: white;
 $jumbotron-padding: 10px;
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,6 +1,9 @@
 class PagesController < ApplicationController
 
   def index
+    @groups = Group.all
+    @people = Person.all
+    @students, @coaches = Membership.all.partition { |membership| membership.name == 'Student' }
   end
 
   def about
--- a/app/views/pages/index.html.haml
+++ b/app/views/pages/index.html.haml
@@ -1,17 +1,27 @@
-.container
-  .row
-    .jumbotron
-      %h1 Rails Girls Groups
-      %h2 Find other Rails Girls students, coaches and project groups!
-      - unless person_signed_in?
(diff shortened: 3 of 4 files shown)


app/controllers/pages_controller.rb
```diff
@@ -1,6 +1,9 @@
 class PagesController < ApplicationController
 
   def index
+    @groups = Group.all
+    @people = Person.all
+    @students, @coaches = Membership.all.partition { |membership| membership.name == 'Student' }
```
til: Nice usage of `partition`!

app/assets/stylesheets/custom/_bootstrap_vars.scss
```diff
@@ -13,5 +13,5 @@ $btn-primary-color: white;
 $btn-primary-bg: $header_red;
 $btn-info-bg: $header_red;
 
-$jumbotron-bg: white;
+//$jumbotron-bg: white;
```
til: I'd rather remove it when not needed, than leaving commented code hanging around.

app/views/pages/index.html.haml
```diff
@@ -1,17 +1,25 @@
...
-      = image_tag "people.png", class: 'landing-image'
-      %h3 Find fellow rubyists
-      %p= link_to "All The Rubyists", people_path, class: 'btn btn-primary'
\ No newline at end of file
+.jumbotron
+  .container.center.section
+    %h1 RoRganize.it
+    %h2 Find other Rails Girls students, coaches and project groups!
+    - unless person_signed_in?
+      %p If you are a student or a coach and not signed up - feel free to register! #{link_to "SignUp", new_person_registration_path, class: 'btn btn-primary', id: 'signup-button'}
```
til: In the navbar, the sign up is written like this: "Sign up". It should be consistent, at least when both appear on the same page.

The haml here could be a bit cleaner when the `%p` was on several lines, then the `link_to` could be on its own line, with an `=` instead of string interpolation.

til:
Great stuff, thanks! :cake:

One might argue that loading all models only to display the counts is inefficient, especially for the often requested home page, but I'd say it is more than fine for now. Unless we have thousands of students and groups, it will probably be not be noticable, and once we have that, it should be easy to change it into only selecting the counts directly from the database.

I'm not merging it yet in case you want to adress the nitpicks (and squash them, yay!) first.

lislis:
fixed the nit-picks, squashed the commits!

PR #311, opened 2015-02-08
Remove non-used mail from Malwine

Malwine:
Dear Rubycorns,
I would like to have this mail removed because I am not using it anymore.
Thanks :cat: 
Lost Rubycorn Malwine


PragTob:
sure thing!

Issue #171, opened 2014-04-07
organize groups by country/city


berlintam:
so that ppl in California don't accidentally sign up for groups in Berlin.

lislis:
I'm closing this, it's kind of part of filters now. Reopen if necessary

PR #319, opened 2015-03-17
Filters

### LINKED ISSUE #182: Split adress string into country, city and address (and maybe office or venue)

lislis:
make group filterable by city and country
closes #182, #171

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -8,6 +8,11 @@ class GroupsController < ApplicationController
 
   def index
     @groups = Group.order :name
+    @groups = @groups.by_country(params[:country]) if params[:country].present?
+    @groups = @groups.by_city(params[:city]) if params[:city].present?
+
+    @group_cities = Group.select(:city).distinct
+    @group_countries = Group.select(:country).distinct
   end
 
   def new
@@ -58,7 +63,7 @@ class GroupsController < ApplicationController
   def group_params
     params.require(:group).permit(:name, :address, :time,
     :picture, :twitter, :contact, :activities, :email, :level,
-    :founded_on, :join_as_coach, :full, :city, :country)
+    :founded_on, :join_as_coach, :full, :city, :country, :street, :zip)
   end
 
 end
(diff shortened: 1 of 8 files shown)


til:
The city filter is a dropdown with all existing cities, right? So if the input to the `by_city` scope can only be from that list, does it really need to match on parts of the word? Or would it be sufficient if the city is simply checked on equality?

lislis:
Ah yes. So before it had to match parts of the city because we didn't check if people also wrote the zip code in there. I then added an explicit zip code field because I made life so much easier. I'll remove the word matching then

PragTob:
What's keeping this from being merged?

lislis:
a thumbs up by a 3rd party?

app/controllers/groups_controller.rb
```diff
@@ -7,7 +7,12 @@ class GroupsController < ApplicationController
   before_action :ensure_member_powers, except: [:index, :show, :new, :create, :destroy]
 
   def index
-    @groups = Group.order :name
+    @groups = Group.where(nil).order :name
```
til: What's the point of `where(nil)` - it doesn't make a difference as far as I can see?

To get a query for all groups without conditions, you could also use `Group.unscoped`, but since there is the order statement appended anyway, this is not necessary.

Also, minor code style remark, I'd use parentheses in `order(:name)`.

app/controllers/groups_controller.rb
```diff
@@ -7,7 +7,12 @@ class GroupsController < ApplicationController
...
+    @groups = @groups.by_country(params[:country]) if params[:country].present?
+    @groups = @groups.by_city(params[:city]) if params[:city].present?
```
til: Nice way of conditionally limiting the results. I like!

app/controllers/groups_controller.rb
```diff
@@ -7,7 +7,12 @@ class GroupsController < ApplicationController
...
+
+    @group_cities = Group.select(:city).distinct
+    @group_countries = Group.select(:country).distinct
```
til: Also nice! :smile:

til:
Maybe you want to try to rebase everything in this PR into one commit? Then it would become a nice single focused commit in the history of the master branch.

Issue #265, opened 2014-09-23
make email field for group required


berlintam:
So that Tam doesn't have to build a messaging system for her pending-student branch...

BriocheBerlin:
Closed with merge #320

PR #320, opened 2015-03-17
Makes contact info and email required for groups.

BriocheBerlin:
+ Group email = required
+ Contact info = required
+ Styles for flash messages updated to pink background and centered
+ All tests pass

We both fields necessary because email is for us (the app maintainers) whereas contact info is for people interested in the group.
Because most groups use googlegroups we didn't change contact info to email field.


til:
:+1: so who wants to deploy it?

til:
#265 can be closed now I suppose?

Issue #137, opened 2014-02-25
Add Emojis for topics :D


LocoKaja:

berlintam:
this is done, oder?

lislis:
YES, closed by 30fbe5ece4ccd73a329e46b6f462ad8

PR #321, opened 2015-03-24
added emoji gem and assets

sareg0:
As per issue #137, we have integrated 'emoji' gem from Steve Klabnik. 
We did not succeed in allowing users to write a word and get the emoji back, so this may be the second step in solving this issue.
If you would like to test it out, you can simple use and image tag and link to the images in 'assets/emoji'. Maybe there is a method but we could not find it. 

Can't wait to use a ridiculous gif if this gets merged :D


PragTob:
What's the status of this? Can we haz merge?

sareg0:
hmm... well there is a bit of a little problem.
Once we added the things, it seems that there are some groups that won't update. Other groups DO update and have the emojis, and then some groups update but DON'T parse the emojis

WAAAT!

lislis:
So the thing is still the renderer @PragTob. The MdEmoji renderer inherits from the Base renderer, but actually we would need it to inherit from the HTML renderer. Ideas? Suggestions?

PragTob:
But.. MdEmoji inherits from the HTML one, no?

`class Render < Redcarpet::Render::HTML` ? https://github.com/elm-city-craftworks/md_emoji/blob/master/lib/md_emoji/render.rb

lislis:
Hmm. But then, why doesn't it work?

PR #325, opened 2015-04-21
attempted to slugify URLs using group name

sareg0:
It is not ready for merge. 

No tests, no real forethought into the rest of the implementation.

This is also for the benefit of Brigitte and Magda and everyone who wanted to work further on the 'slug' issue :)

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -39,8 +39,9 @@ class Group < ActiveRecord::Base
   mount_uploader :picture, PictureUploader
   geocoded_by :location
   after_validation :geocode
+  after_save :update_slug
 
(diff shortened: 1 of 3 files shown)


berlintam:
I don't know if you're interested in using a gem for this, but if you are, this one's pretty good! https://github.com/norman/friendly_id

sareg0:
We did discuss it. We were just looking to get some experience implementing it too, but have determined that the work to do the backwards compatibility for making the old URLs still redirect would be too hard by hand, so the gem will probably the best way to move forward... If we don't want to reinvent the wheel ;D

BriocheBerlin:
Done in other branch, see #330

Issue #327, opened 2015-04-28
blog posts displays date of creation, not of publication


lislis:

lislis:
closed by 506b23da0ee6e57279

PR #334, opened 2015-05-14
Add publication date to posts

berlintam:
I'm so lost at the moment. This is so solve the bug https://github.com/rubycorns/rorganize.it/issues/327. I decided not to use the `updated_at` column as the publication date, because what if someone updates an old post!? It's publication date would change! That would just be madness. So I created a new `publication_date` column for posts and am trying to use that.

In post#update, I'm trying to set a post's publication date if it was previously a draft, and is no longer. That's the `!@post.draft && @post.draft_changed?` part. 

Then the `@post.set_publication_date!` method is called, which should set the post's publication date. 
IT DOES. I can see it happening with all of my crazy ramblings in the code there where I look to see if it's been set. 

However, once the post is reloaded, it loses its publication date! why?! For some reason it's not being saved.... and I can't figure it out, help please!

ps - in my opinion, and in the internet'z opionion, `update_attribute(:publication_date, Date.today)` should save the object... which only furthers my confusion. My guess is it's being overwritten somewhere. But I don't know where.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,6 +96,7 @@ group :development, :test do
   gem 'launchy'
   gem 'minitest'
   gem 'rspec-its', '~> 1.0.1'
+  gem 'timecop'
 end
 
 # Use ActiveModel has_secure_password
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -269,6 +269,7 @@ GEM
     thor (0.19.1)
     thread_safe (0.3.4)
     tilt (1.4.1)
+    timecop (0.7.3)
     timers (1.1.0)
     treetop (1.4.15)
       polyglot
@@ -334,6 +335,7 @@ DEPENDENCIES
   spring
   spring-commands-rspec
   sqlite3
+  timecop
   turbolinks
   uglifier (>= 1.3.0)
   will_paginate
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -27,6 +27,8 @@ class PostsController < ApplicationController
     @post.draft = post_params_draft
     @post.person = current_person
 
+    @post.set_published_on! if !@post.draft
+
     if @post.save
       redirect_to @post, notice: post_created_notice
     else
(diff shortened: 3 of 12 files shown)


til:
One thing that looks suspicious to me is that the publication_date column in schema.rb is of type "time". What is that - does it maybe mean time without date? E.g. created_at has the type "datetime". Anyway, I wanted to suggest a change, which may or may not fix this problem: in rails there is the convention to call attributes that contain date and time something with the postfix "_at". E.g. "created_at", but the convention also makes sense for custom attributes, such as "finished_at", "published_at", etc. Then there is another, lesser used, but still useful, convention - to postfix columns that only contain a date, but no time, with postfix "_on". And in this case, we really only are interested in the date, and do not want to store a time with it, or do we? So we could call the column "published_on", and also change it to be of type "date".

berlintam:
oh man... that totally worked. thanks Til! I was trying to do a datetime, but in my migration I wrote date_time instead of datetime... and I guess that did something different. 

Thanks! now tests :tongue:

til:
Wow, it's nasty that datetime and date_time silently do different
things, I didn't know that. Anyway, glad to hear it works now!

berlintam:
I admit that there could be more tests, but I think that the most important thing (the publication date being set in the controller) is tested, and I'm a bit tested out now. If anyone wants more, just let me know.

lislis:
@berlintam can you squash and rebase?

berlintam:
k done!

lislis:
can you maybe squash into one commit? For a nicer commit history?

til:
But also squash! The commits together into one! Please ...

And then `git push -f` to this branch.

berlintam:
you kids are crazy.

berlintam:
squish

lislis:
awesomesauce!

berlintam:
was this deployed? If so, my rake task has to be run on the server....

til:
Not by me, maybe by Lisa?

lislis:
I deployed it. But didn't run the rake task. Can we do it with the
shelly console?

PR #335, opened 2015-05-20
add 'cache' line in travis.yml file

sareg0:
As per @lislis's request, we added this line to the travis.yml file that will cache our bundle, and make our tests run much faster...Correct me if I'm wrong Lisa :)


PragTob:
most definitely, great thing to have around, waiting for the build to start and finish

PR #338, opened 2015-05-26
adds willing to coach checkbox

sareg0:
as per the issue in wish @alicetragedy outlined the need for a coach list, we are starting by adding a checkbox to the user's profile, that indicated their willingness to coach.

This will be built out to a bigger feature, but for now the idea is that each user in the project group app could indicate that they would like to coach a workshop. We will then use this to populate a list that people can search if they are in need of coaches.

Of course people registered as coaches will already be listed as such.

### DIFF
--- a/spec/features/workshop_coach_spec.rb
+++ b/spec/features/workshop_coach_spec.rb
@@ -1,6 +1,7 @@
 require 'spec_helper'
 
 feature 'workshop coach' do
+  before { login_as_user }
 
   scenario 'see workshop coach text' do
     visit_profile_page_of_a_workshop_coach
@@ -11,13 +12,13 @@ feature 'workshop coach' do
 
   def login_as_user
     person = create(:person)
+    visit new_person_session_path
     sign_in person
   end
 
   def coach_text
     'I am willing to coach at Rails Girls workshops'
   end
-
 
   def see_workshop_coach_text
     expect(page).to have_content(coach_text)
@@ -28,8 +29,6 @@ feature 'workshop coach' do
   end
 
   def visit_profile_page_of_a_workshop_coach
-    visit new_person_session_path
-    login_as_user
     person = create(:person, first_name: 'Sara', workshop_coach: true)
     click_link 'People'
     click_link 'Sara'
--- a/spec/features/workshop_coach_spec.rb
+++ b/spec/features/workshop_coach_spec.rb
@@ -1,0 +1,44 @@
+require 'spec_helper'
+
+feature 'workshop coach' do
+
+  scenario 'see workshop coach text' do
+    visit_profile_page_of_a_workshop_coach
+    see_workshop_coach_text
+    visit_profile_of_normal_person
+    not_see_workshop_coach_text
+  end
+
+  def login_as_user
+    person = create(:person)
+    sign_in person
+  end
+
+  def coach_text
+    'I am willing to coach at Rails Girls workshops'
+  end
+
+
+  def see_workshop_coach_text
+    expect(page).to have_content(coach_text)
+  end
+
+  def not_see_workshop_coach_text
+    expect(page).not_to have_content(coach_text)
+  end
+
+  def visit_profile_page_of_a_workshop_coach
+    visit new_person_session_path
+    login_as_user
+    person = create(:person, first_name: 'Sara', workshop_coach: true)
+    click_link 'People'
+    click_link 'Sara'
+  end
+
+  def visit_profile_of_normal_person
+    person = create(:person, first_name: 'Foobar', workshop_coach: false)
+    click_link 'People'
+    click_link 'Foobar'
+  end
+
+end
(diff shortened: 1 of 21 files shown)


berlintam:
I think this is a great start!

A few comments: Laura's issue is about ppl willing to coach at, as I understood it, RG workshops, potentially in other cities. A checkbox called 'willing to coach' doesn't really portray this. Especially in the context of an app for rails girls project groups... It more makes it seem like I, as a student, can sign up for a group, and can simultaneously agree to coaching that group occasionally.... This then kind of blurs the lines between student and coach... which is totally fine... but in the context of our app is a bit confusing. (which may result in such things like WHO AM I!? WHAT IS EVEN A COACH?!)

I suggest just re-naming it, to something like "willing to coach at workshops" or something along those lines... and perhaps some explanatory text... like, this information will be shared with others, specifically rails girls chapters.

You know what i mean?

alicetragedy:
@berlintam I agree with your comments. 
tbh I also see a difference between coaches at study groups and coaches at RG events (but I think in that respect every study group is different).
I will make sure that I outline the needs of the "RG workshop coach" feature in the issue in more detail so as not to create confusion :)

sareg0:
that makes total sense, and it is something that came to mind. We can fix that.

I'm gonna mark this a hackday issues since it's nice and easy to take a swing at. :)

PragTob:
this seems stale, who is working on it?

berlintam:
@sareg0

sareg0:
this is an effort to create a coach list for rails girls organisers to take a look at if they are in need. It is mainly because other things came in the way that it was been left. I think that tests needs to be written, and the text on the checkbox needs to be fixed.

berlintam:
let me know if you need help @sareg0 !

sareg0:
Ah! I didn't get any further with this, but I would like to. However, if anyone does take over before next meeting that's totes fine.

til:
I agree with the comments, and just want to add that obviously this was
not meant to be the final, sufficient, and correct way to satisfy the
feature request, but rather just a first stab at something, which could
go in the right direction.

So, to make it more clear that the checkbox expresses the willingness to
coach at RG workshops, but not project groups, we could name the field
"willing_to_coach_at_workshops". Or
"willing_to_coach_at_railsgirls_workshops". Or, to avoid it being overly
long: "workshop_coach". Actually I think the last one fits well. What do
you think?

sareg0:
I just added the stuff I worked on on Tuesday, but I still need to write an acceptance test for this, which I don't know how to do. @PragTob said he could help with this, so I will see about doing it and finishing next Tuesday

sareg0:
AAAAAAHHH!! The acceptance test is here! Can someone please review the magic, and tell me if my spells need tweaking? Many thank yous

PragTob:
@sareg0 could you also already attempt the merge so it can be automatically merged? :)

spec/features/workshop_coach_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def login_as_user
+    person = create(:person)
+    sign_in person
+  end
+
+  def coach_text
+    'I am willing to coach at Rails Girls workshops'
+  end
```
PragTob: This is best practice wise better as a constant, but it won't change behavior :)

spec/features/workshop_coach_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    expect(page).to have_content(coach_text)
+  end
+
+  def not_see_workshop_coach_text
+    expect(page).not_to have_content(coach_text)
+  end
+
+  def visit_profile_page_of_a_workshop_coach
+    visit new_person_session_path
+    login_as_user
```
PragTob: it'd be good if the whole login as a user thing would be a separate step - like this it is implicitly included in this step and should we change the order of the visit profile path then the specs would break

--> separate "I'm logged inas a suer step" at the beginning of the spec or a s a before :each or something

til:
:rocket: :cake: :dancer: !

spec/features/workshop_coach_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    person = create(:person, first_name: 'Sara', workshop_coach: true)
+    click_link 'People'
+    click_link 'Sara'
+  end
+
+  def visit_profile_of_normal_person
+    person = create(:person, first_name: 'Foobar', workshop_coach: false)
+    click_link 'People'
+    click_link 'Foobar'
```
PragTob: just icing on the cake, but the "visit profile of person" functionality could be extracted into a method :)

PragTob:
minor comments otherwise, everything is great :+1:

spec/features/workshop_coach_spec.rb
til: True, a constant would better communicate that this is, uhm, a constant string. I suggested to make it a method though - I usually do it like this, and would actually prefer it over a constant, because using bare words feels so rubyish imho. and COACH_TEXT would just be screaming ...

sareg0:
@PragTob , wdyt?

til:
Rather close this and merge #351 instead?

PR #341, opened 2015-05-31
actually USE published_on date

berlintam:
Ummmmmm turns out I totally did all the backend for for using the `published_on date`, (see https://github.com/rubycorns/rorganize.it/pull/334) but I didn't actually USE it anywhere. This PR handles that.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -7,8 +7,7 @@ class PostsController < ApplicationController
   require 'will_paginate/array'
 
   def index
-    @published_posts = Post.where(draft: false)
-    @published_posts.order(published_on: :desc)
+    @published_posts = Post.published
     page = (params[:page] || 1).to_i
     @published_posts = @published_posts.paginate(page: page, per_page: 20)
   end
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -7,9 +7,10 @@ class PostsController < ApplicationController
   require 'will_paginate/array'
 
   def index
-    @posts = Post.all.order(created_at: :desc)
-    @page = (params[:page] || 1).to_i
-    @posts = @posts.paginate(page: @page, per_page: 20)
+    @published_posts = Post.where(draft: false)
+    @published_posts.order(published_on: :desc)
+    page = (params[:page] || 1).to_i
+    @published_posts = @published_posts.paginate(page: page, per_page: 20)
   end
 
   def show
@@ -44,7 +45,7 @@ class PostsController < ApplicationController
     end
       
     if @post.update(post_params)
-      redirect_to @post, notice: 'Post was successfully updated.
+      redirect_to post_path(@post), notice: 'Post was successfully updated.
       All efforts, nomatter how small, deserve cake.'
     else
       render action: 'edit'
--- a/app/views/dashboards/show.html.haml
+++ b/app/views/dashboards/show.html.haml
@@ -57,19 +57,16 @@
             %li= link_to fa_icon('plus', text: 'create new post'), new_post_path
           - if @published_posts.any?
             %h2
-              %small Latest Published Post:
+              %small Latest Published Posts:
               %br/
-            = link_to @published_posts.first.title, post_path(@published_posts.first.id)
-            = link_to fa_icon('edit', text: 'edit'), edit_post_path(@published_posts.first.id)
-            = link_to fa_icon('trash-o', text: 'delete'), post_path(@published_posts.first.id), method: :delete, data: { confirm: "You really want to trash that?" }
+            = render partial: 'posts', locals: { posts: @published_posts.limit(2) }
 
           - if @unpublished_posts.any?
             %h2
               %small Unpublished posts:
-            - @unpublished_posts.each do |post|
-              = link_to post.title, post_path(post.id)
-              = link_to fa_icon('edit', text: 'edit',), edit_post_path(post.id)
-              = link_to fa_icon('trash-o', text: 'delete'), post_path(post.id), method: :delete, data: { confirm: "You really want to trash that?" }
+              %br/
+            = render partial: 'posts', locals: { posts: @unpublished_posts }
+
     .col-md-4.col-sm-6
       .panel.panel-default.panel-image
         .panel-heading.clearfix
--- a/app/views/dashboards/show.html.haml
+++ b/app/views/dashboards/show.html.haml
@@ -55,19 +55,21 @@
           %ul.dashboard-list
             %li= link_to fa_icon('list', text: 'all posts'),posts_path
             %li= link_to fa_icon('plus', text: 'create new post'), new_post_path
-          - if @posts.exists?
+          - if @published_posts.any?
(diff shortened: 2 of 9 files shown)


app/views/posts/index.html.haml
```diff
@@ -2,14 +2,15 @@
   .row
     %h1.col-sm-12.page-header Corporate Blog
     .col-sm-8
-      - @posts.each do |post|
-        - if post.draft == false || (current_person && current_person.has_role?(:admin))
-          = render partial: 'post', object: post
-      - if @posts.empty?
+      - if @published_posts.any?
+        - @published_posts.each do |post|
+          - if post.draft == false || (current_person && current_person.has_role?(:admin))
```
PragTob: would be nice if we'd have a `post.draft?` method here to avoid the comparison (actually through active record magic we have this method, but we might want to write our own/test it)

also the current_person && role admin would be nice to have as a helper not to repeat it

app/controllers/posts_controller.rb
```diff
@@ -7,9 +7,10 @@ class PostsController < ApplicationController
   require 'will_paginate/array'
 
   def index
-    @posts = Post.all.order(created_at: :desc)
-    @page = (params[:page] || 1).to_i
-    @posts = @posts.paginate(page: @page, per_page: 20)
+    @published_posts = Post.where(draft: false)
+    @published_posts.order(published_on: :desc)
```
til: Does this line do anything? Looks like the return value is simply discarded, I don't think this actually changs the order of `@published_posts`.

Since this logic is used both here and in the dashboard controller, maybe think about moving it into a scope (or just a class method) of the model, so that you can just call `Post.draft` and `Post.published`, and get back the filtered records with correct ordering.

The `paginate` call on the other hand I'd leave in the controller, because it sounds more of a concern of the display logic than the model itself. E.g one controller might want to paginate into smaller pages, another one into larger ones.

app/views/dashboards/show.html.haml
```diff
@@ -55,19 +55,21 @@
...
+            = link_to fa_icon('edit', text: 'edit'), edit_post_path(@published_posts.first.id)
+            = link_to fa_icon('trash-o', text: 'delete'), post_path(@published_posts.first.id), method: :delete, data: { confirm: "You really want to trash that?" }
+
+          - if @unpublished_posts.any?
+            %h2
+              %small Unpublished posts:
+            - @unpublished_posts.each do |post|
+              = link_to post.title, post_path(post.id)
+              = link_to fa_icon('edit', text: 'edit',), edit_post_path(post.id)
+              = link_to fa_icon('trash-o', text: 'delete'), post_path(post.id), method: :delete, data: { confirm: "You really want to trash that?" }
```
PragTob: This looks fairly similar to what the published post above does, having a partial for rendering a single post seems like a good idea :)

app/controllers/posts_controller.rb
PragTob: ahh good catch @til ! :)

I'm not sure but maybe `order!` is a thing?

PragTob:
Some comments + le tasts are le failing!

Greetings from Toni!

app/views/posts/index.html.haml
berlintam: actually i think that whole thing can just be removed. `@published_posts` already checks to see if the post is a draft or not.

app/controllers/posts_controller.rb
berlintam: ah good idea!

app/views/dashboards/show.html.haml
berlintam: ya, we need a big refactor w/ partials in general

berlintam:
thanks for the comments! I shall fix :)

berlintam:
done

spec/models/post_spec.rb
```diff
@@ -2,8 +2,14 @@
 
 describe Post do
 
-  it { is_expected.to belong_to(:person) }
+  describe 'scopes' do
+    specify do
+      expect(Post.published.to_sql).to eql "SELECT \"posts\".* FROM \"posts\"  WHERE \"posts\".\"draft\" = 'f'  ORDER BY \"posts\".\"published_on\" DESC"
+      expect(Post.draft.to_sql).to eql "SELECT \"posts\".* FROM \"posts\"  WHERE \"posts\".\"draft\" = 't'  ORDER BY \"posts\".\"created_at\" DESC"
```
PragTob: this might be me but I don't fancy testing the SQL statements that are generated. For me those are a lower level detail that I don't care about (most times) :) Also it seems a bit brittle.

I'd rather test given data x in the database, are exactly those entities in the scope that I'd expect to be there?

berlintam: ya, I think it all comes down to personal preference. I know at work we test it this way, b/c testing that an object is present in the database required making that object, which ends up slowing down tests, correct?

I dunno thou, since this is a learning app, perhaps your way is a bit more understandable? & it's not like we have a huge & slow test suite to worry about... (yet...)

wdyt?

til: You test it at work like this, comparing against the exact generated SQL? Can I have a word with these people please? ;-)

True, actually creating an entry in the database takes a little bit of time, and also there is the argument that you shouldn't test code that you do not own. In this case, when testing something by creating an entry in the database, we would in parts also test the behaviour of ActiveRecord, which we do not own. But testing the generated SQL does so too.

I've found that for code like this, which heavily depends on interacting with the database, it's useless trying to avoid the database in the tests. Just make sure you don't create hundreds of entries or queries in one test, and the speed should remain fine.

I've pushed a commit to this branch with an example of specs for the scopes that use the database. Choose which variant you find more readable ...

berlintam: thanks @til! can we keep both in there? I always learn something from looking at your fancy rspec (z.B. `contain_exactly`)

Althou some ppl also say u don't need to test scopes at all..... what do you think about that?

til: Well yes, I guess we can leave both in there. Maybe add a comment that
those are examples of testing the same thing in different ways?
Otherwise it might become too confusing. I wonder what makes a better
learning app - one that has all kinds of different things mixed
together, or one that strictly obeys consistency and tries to do
everything in the best possible way? If only there was _one_ best
possible way ... ok, forget about the consistency then.

I guess above all, our app should encourage experiments, contributions
and improvements. So let's leave both specs, and once someone sees it as
an improvement to remove one or both, let them be allowed to do so.

Not testing scopes at all - sure, that's possible too. Most of the time
they are so simple that it is obvious what is happening anyway. I guess
in a normal, typical project of mine I would also not test them, or
maybe test only the most crucial aspect, which in this case would be
that published really excludes the posts marked as draft. But maybe I
wouldn't spec the Post.draft scope, and also not the ordering of
Post.published. Testing ordering in the database requires always a
little effort, because the spec must make it unlikely that it passes due
to an accidentially correct order.

However, what I would do for sure in a typical project, add a high-level
feature spec that at least once requests the page where each scope is
used. So that removing the scope definition from the model would
definitely generate a spec failure.

Issue #279, opened 2014-11-20
send email to group members when a new topic is suggested


berlintam:
It just occurred to me that I'd like to talk about servers one day. So I was going to put it in the topics section of our app. But then I decided not to, because I figured I'd probably forget I'd put it there, so there isn't much point... which made me think if an email was sent out to all members of the group whenever a new topic is suggested, maybe topics won't be forgotten. 
Or maybe the day before the project group, an email can be sent out to all members reminding them to come to the project group (cough cough @FrauBienenstich cough cough), and also telling them about potential topics. 
That way maybe we can play with workers and sidekiq and stuff too! (which sure, sounds like overkill, but I use this stuff at work all the time, I'd like to have a safe place to play around with it a bit)

thoughts?

PR #344, opened 2015-06-09
Notifications

lislis:


app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def new
+    @notification = Notification.new
+  end
+
+  def create
+    @notification = Notification.new(notification_params)
+
+    @notification.person = Person.find(params[:notification][:person_id].to_i)
+    @notification.group = Group.find(params[:notification][:group_id].to_i)
```
til: These two lines don't look as if they were necessary, this should theoretically all be possible to be done by the model instantiation. Are you sure `params[:notification][:person_id]` and `params[:notification][:group_id]` are set to anything here? Wouldn't they need to be permitted explicitely first?

app/models/person.rb
```diff
@@ -69,6 +70,14 @@ def member_of?(group)
     !!memberships.find_by(group_id: group.id)
   end
 
+  def has_notifications?
+    notifications.present?
+  end
+
+  def list_notifications
+    notifications
+  end
```
til: Not sure I see the benefit of this method over simply `notifications`, did you plan to extend this later or why was it added?

app/views/notifications/index.html.haml
```diff
@@ -0,0 +1,11 @@
+.container.student-list
+  %h1.page-header Notifications Overview
+  .row
+    .col-md-8
+      %ul.list-group
+        - if @notifications.present?
```
til: Maybe swap these two lines to avoid an empty `ul` when there are no notifications?

app/views/notifications/_form.html.haml
```diff
@@ -0,0 +1,14 @@
...
+  = render 'shared/error_messages', object: f.object
+  .form-group
+    = f.label(:person_id, "User")
+    = f.select(:person_id, options_from_collection_for_select(Person.all, :id, :name))
+  .form-group
+    = f.label(:group_id, "Group")
+    = f.select(:group_id, options_from_collection_for_select(Group.all, :id, :name))
+  .form-group
+    = f.label(:content, "Content")
+    = f.text_field :content, placeholder: "What what?", class: 'form-control'
```
til: :laughing:

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    if @notification.save
+      redirect_to notifications_path, notice: 'A new notification was created. This calls for cake!'
```
berlintam: :+1:

db/schema.rb
```diff
@@ -63,6 +63,15 @@
   add_index "memberships", ["group_id"], name: "index_memberships_on_group_id"
   add_index "memberships", ["person_id"], name: "index_memberships_on_person_id"
 
+  create_table "notifications", force: true do |t|
+    t.integer  "person_id"
+    t.integer  "group_id"
+    t.datetime "viewed_at"
```
berlintam: what exactly does this `viewed_at` thing do? Does it show the person who sent the notification the time it was viewed at by the receiver? or does it just tell the receiver what time they themselves viewed it at? If it's the first one, I'd hardcore argue not to have it. This is my least favourite feature of facebook. It makes me feel like I'm being stalked.

lislis:
Thanks for the feedback people! Super much appreciated! This is the weird first step of something that is not yet fully fleshed out in my brain. So here are bits of my thoughts:  
- This is the first step in having group membership approval. If someone wants to join your group, they hit a button and the group members get a notification.
- This is also useful to notify the applying person that they membership has been approved or rejected.
- I want admins to have the possibility to send notifications to a person or a group or basically everyone. Like when we change requirements for group information and say: "Please update your group's info, contact email is now required"
- We can also send out notifications when the group itself gets updated or we release a blog post or other things.

I'm not sure how this all should work in detail. I'll iterate over it in the next days, feel free to make suggestions.

app/controllers/notifications_controller.rb
lislis: I have no idea. Will check

app/models/person.rb
lislis: Probably. I don't know yet...

app/views/notifications/index.html.haml
lislis: YES!

db/schema.rb
lislis: So if  `viewed_at` is nil the notification is active and users will get a little bubble that says `1`. Once the notification is viewed (basically not nil and I use a datetime here) the notification bubble vanishes. Then I want  to use the date of `viewed_at` to delete all notifications that are older than a month or so. Does that make sense? I guess naming was poor on my side

til: viewed_at sound to me like a very intuitive, rails-ish name for
this. It's common to use dates with _at postfix as boolean indicators
that something happened. E.g. deleted_at to keep deleted records in a
table, and only display "WHERE deleted_at IS NULL" by default.

berlintam: ah ok, got it, no makes sense. or maybe `viewed` as a boolean?

berlintam:
fyi, not sure if this is important at this stage, but when a person joins a group, that group gets sent an email. (https://github.com/rubycorns/rorganize.it/blob/master/app/mailers/person_mailer.rb) I tested it out tonite and it does work for realz. We just don't get a notification b/c we had the google group as our email (this is why we also don't get travis failure emails). I have since changed this to be my email address (I'll forward it to the group if we get sent an email). 

We should create a notifications@rubycorns.club email address (or something like that), and forward emails that get sent to that to the google group. 

so, not sure if this is important info for this PR, but... there it is.

berlintam:
oh, it looks like this issue is also related to this: https://github.com/rubycorns/rorganize.it/issues/279

lislis:
@til do you have time to look at what I did here?

app/models/person.rb
```diff
@@ -109,6 +110,10 @@ def member_of?(group)
     !!memberships.find_by(group_id: group.id)
   end
 
+  def unread_notifications
+    notifications.select { |notification| notification.viewed_at == nil }
+  end
+
```
til: Does this work as expected? This is not a rethoric question - I am actually unsure whether it would work as expected or not, and always return all notifications instead. `select` is normally a method of ruby's Enumerable, but in this case it is a method of ActiveRecord associations (because `notifications` is a so-called association of Person here).

If it works as expected - do you know how it works exactly? Does it first fetch all notification rows from the database and the loops over them in ruby? Or does it only fetch the rows that it needs? How could you find out? Either try it in pry and look at the generated sql, or maybe even add a little model spec that runs just the `unread_notifications` method in isolation.

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,73 @@
+class NotificationsController < ApplicationController
+
+  before_action :authenticate_person!
+
+  def index
+    @notifications = Notification.all
```
til: This shows all notifications of all users - is this intended? What are notifications exactly? ... reads the schema and the test file ... slowly understands - ah, notifications are like system wide broadcast messages? And can only be entered by admins? I see. Maybe add some details to the scenario of the feature spec that makes that more clear?

In any case, at least this list here should be sorted by date I guess.

spec/features/notifications_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+feature 'Notifications' do
+
+  subject { page }
+
+  let(:person) { create(:person) }
+
+  before do
+    visit new_person_session_path
+    sign_in person
+    person.update_attribute(:admin, true)
```
til: There must be a more elegant way to create a user in the spec that is admin ...

app/models/person.rb
berlintam: I think you could turn this into a scope instead à la `notifications.where(viewed_at: nil)`-ish

app/views/people/show.html.haml
```diff
@@ -60,6 +60,9 @@
         %p= markdown(@person.working_on)
       - elsif @person == current_person
         %p Why not tell the world what you're working on?
+    .col-md-12
+      - @person.notifications.each do |note|
```
berlintam: I think that these notifications should all be on separate lines instead of all on the same line, otherwise they'll be quite difficult to read. I think it would also be nice to have the ability to delete them from the show page, not just from the dropdown.

spec/features/notifications_spec.rb
berlintam: indeed there is :) https://github.com/rubycorns/rorganize.it/blob/master/spec/models/person_spec.rb#L146

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      if @notification.update_attributes notification_params
+        redirect_to root_path, notice: 'Notification was updated. Let\'s have some cake.'
+      else
+        redirect_to root_path, notice: 'Could not update notification. Womp womp.'
+      end
+    end
+  end
+
+  def show
+    # notifications = Notification.where(notification_id: @notification.id)
```
berlintam: you can probably just delete this, hey? do we still need the edit & show methods at all? I wonder if rails like forces you to have them just for consistency's sake....

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def new
+    @notification = Notification.new
+  end
+
+  def create
+
+    group_id = params[:notification][:group].to_i
+    group = Group.find(group_id) unless group_id == 0
```
berlintam: I don't think you need the `unless group_id == 0` part. That shouldn't happen....

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    success = true
```
berlintam: You shouldn't need this line. If the save is successful, it'll be true, otherwise it'll be false.

berlintam:
does a person or a group have precedence? Or can you choose both?

screen shot 2016-01-12 at 21 32 33

berlintam:
The styling on the show page is a bit confusing now: it looks as though the delete button here is for the notification, but it's for one's account! :scream: 

screen shot 2016-01-12 at 21 33 43

There's also some wonky things happening on the /notifications page

screen shot 2016-01-12 at 21 34 32

berlintam:
woot woot! looks awesome - it'll be cool to have this up & running!! Just thing of all the notifications we can send...... mwahahahahaaaaa

app/views/people/show.html.haml
lislis: tbh. I don't even know why that is there... I wanted notifications to only be in the topbar

app/controllers/notifications_controller.rb
lislis: :ok:

app/controllers/notifications_controller.rb
lislis: I removed them

lislis:
I'm closing this. After talking to @til we'll take a different approach and start from scratch.

Issue #339, opened 2015-05-30
learning resources section


berlintam:
have a section that lists learning resources which come RECOMMENDED by the fabulous rubycorns!

lislis:
Is this meant to be part of rorganize or of the rubycorns.club? in case of the first one, please elaborate on how you image it

berlintam:
I guess I was thinking for rorganize it. Or maybe we can have a 'getting started' section, and list off things people can do in a project group, like 'work on the following tutorials which we think we really good' and then list them, or 'contribute to the following open source project which are beginner friendly' and then list them etc....

It would just be a static page which we, as admins, could add to. 

Or perhaps we can expand the groups show pages a bit, and individual groups can add some of their favourite learning resources, which other ppl can then see and use for themselves.

til:
I like the idea of having a space for project group specific learning
materials. Every group is different, and learns different things, some
don't even do ruby but e.g. clojure, so it'd feel wrong to me for us to
impose a global list of recommendations on everyone.

It could be a simple textarea field per group, that is only displayed on
the groups show page when filled out. Of course with emoji and autolink
support!

It would be interesting to see what the various groups put in there! I'd
put in for the rubycorns: http://ruby-for-beginners.rubymonstas.org/ ,
https://www.railstutorial.org/ and http://guides.rubyonrails.org/ (as a
general recommendation for intermediate-ish people).

berlintam:
:+1:

berlintam:
closed w/ #345

PR #345, opened 2015-06-14
Learning resources

berlintam:
This closes issue #339

Along the way, I also re-arranged the order of fields in the groups form. I think before we were just adding whatever new field to the end of the form, so things were a bit mixed up. I think it makes more sense now, but let me know if I'm way off base. 

I also added some text to clarify the difference between the email field and the contact field.

I also noticed that the form made it look as though city & country are required (they had *'s beside them). This however was not the case for realz. I added validations to these two fields & fixed the tests. Was there a reason why we were presenting them like they were required but weren't actually making them required?

I think once this is merged we should send an email to all groups asking them to update their information.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -83,7 +83,6 @@ ActiveRecord::Schema.define(version: 20150614105629) do
     t.string   "picture"
     t.string   "twitter"
     t.text     "working_on"
-    t.boolean  "willing_to_coach"
   end
 
   add_index "people", ["email"], name: "index_people_on_email", unique: true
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -82,6 +83,7 @@
     t.string   "picture"
     t.string   "twitter"
     t.text     "working_on"
+    t.boolean  "willing_to_coach"
```
til: This was added by accident: probably because you ran the migration from the branch where this column was added on your local dev system. After that, the column was in your dev database and ended up in here. I suggest to manually remove the line again, before this PR gets merged. Alternatively you can checkout master, run rake db:reset there, then switch to this branch and run the migrations - this should have the same effect.

berlintam:
moss

til:
Looks good, thanks! Also, good idea to clarify what the contact fields are for and who can see what.

til:
Merged, but I'm leaving the deploy to whoever feels like deploing.

lislis:
@berlintam you're on fire!

berlintam:
fire

lislis:
:+1:

PR #349, opened 2015-06-30
User experience

### LINKED ISSUE #347: improve user experience
Maybe it would be a good idea to give new users of our website a bit of a tour. At the moment, users sign up and are just brought back to the root page, with no explanation. 

Maybe we could have a 'new user landing page' or something, that sort of explains what to do. Like, go join a group, make a group, put what you're working on in your profile, read our blog, etc...

Or perhaps a email to new users would be enough, one that provides links to helpful parts of our site to get ppl started. 

wdyt? 

ps - gifs should be involved in as many places as possible.

berlintam:
OMG SO PRETTY closes #347


berlintam:
I was thinking as an addition to this, maybe we could have another picture & explanation to tell ppl to fill in their profile. @BriocheBerlin do you want to take a stab at that? Or shall we tackle it together at the next cornz?

app/controllers/registrations_controller.rb
```diff
@@ -15,6 +15,10 @@ def destroy
     Please consider sending the programmers of this app one last piece of cake.'
   end
 
+  def after_sign_up_path_for(resource)
+    '/welcome'
```
til: Can we use the URL helper `welcome_path` instead, as it is the true and proven way in rails?

berlintam: sure, I was wondering about that as well actually, like... what's the right way to do things. I guess `_path` it is!

app/views/pages/welcome.html.haml
```diff
@@ -0,0 +1,74 @@
...
+  .row.make-cake-section
+    .col-sm-6.additional-suggestions
+      .row
+        .col-sm-offset-3.col-sm-7
+          %h2
+            Then
+          %ul
+            %li
+              read our
+              %a{href: '/posts'} blog
```
til: Same here - `posts_path`?

What's the point? It's more general. For example: if sometimes later we decide to translate the page, and all URLs change from `/posts` to `/de/posts`, then using helpers this change will work automatically.

app/controllers/registrations_controller.rb
til: It is. You can use `rake routes` if unsure which path helpers exist. The
left column is the name of the helper method, just append "_path" or
"_url" to it.

app/views/pages/welcome.html.haml
berlintam: makes sense, I shall change. Thanks @til!

PragTob:
I"m :+1: on shipping this early and then later on we can still make improvements :) So I'm just gonna hit the big GREEN button :)

berlintam:
ok! thanks @PragTob. did u deploy?

PragTob:
nope

PR #350, opened 2015-07-07
Add profile info to Welcome page

BriocheBerlin:


app/assets/stylesheets/index.html
```diff
@@ -0,0 +1,34 @@
...
+    .changeonhover:hover {
+        background-color: blue;
+        width: 600px;
+        height: 300px;
+    }
+</style>
+</head>
+<body>
+<div>Hover me! Hover Me!</div>
+</body>
```
til: Was this file added on purpose?

til:
I restarted the travis build, looks like travis had some hiccups.

Gemfile.lock
```diff
@@ -342,6 +342,3 @@ DEPENDENCIES
   turbolinks
   uglifier (>= 1.3.0)
   will_paginate
-
-BUNDLED WITH
```
berlintam: oops, sorry @BriocheBerlin I told u to commit this but you shouldn't have actually. @PragTob what do we do now?

til: Just revert this change to this file, and push that as new commit to this branch?

Bonus points for squashing it, but can be done later.

Oh, and run a `gem install bundler` locally to make sure you have the most recent one installed. Then you should be safe from getting this change applied automatically again, I think there were a few recent versions of bundler that always changed this section, but the most recent one only does it if the Gemfile.lock also changed elsewhere.

app/assets/stylesheets/index.html
PragTob: I don't think so :)

Gemfile.lock
PragTob: What @til said :)

I can also do the squashin :)

berlintam: thanks @til & @PragTob!

PR #352, opened 2015-07-12
Omniauth2

FrauBienenstich:
do et


berlintam:
shellycloud should also be properly set up now too

til:
Looks good to me besides the minor comments, but to be honest I can't give useful feedback regarding the functionality, because I have no experience with omniauth, and I'm too lazy right now to actually try it out. 

An acceptance spec that tests this functionality would be ubercool of course, but I guess it would be extra-hard, because you'd need to find a way to mock out the GitHub response somehow ...

PragTob:
@til yep an acceptance spec would be great.. thought about it, but you'd need to be logged in with github so we'd have to try and capture that and overall that's.. maybe too much hassle. I talked with @FrauBienenstich how we could at least test the controller etc. :)

til:
I understand totally. Yes it would be a bit of a hassle.

I assume that vcr can be used for the capturing part. But there are more
rewarding situations where to start using vcr, e.g. an API where also
content is retrieved and parsed from, not just an API that is used for
authentication.

PragTob:
the worst thing is we couldn't even use vcr for the initial auth as we physically have to go to github and github redirects us back, but we can test our controller by doing a redirect like github would :)

til:
Ah I see - even worse indeed.

berlintam:
keys have been fixed on da server

berlintam:
:shipit: ?

berlintam:
@PragTob can u give a final thumbs up on this?

PragTob:
I erm.. derp didn't see this. I'll go through the code and just assume you checked it all working and stuff and then I'll merge and we fix it later :)

PS: Feel free to ping me if I don't respond to issues within ~3 days or so. Bad tobi.

app/models/person.rb
```diff
@@ -54,6 +63,12 @@ def name
     full_name
   end
 
+  def name=(string)
+    names = string.split(' ')
+    self.first_name = names.first
+    self.last_name = names.last
```
PragTob: we should log an issue here that this disregards second names etc.

config/secrets.yml
```diff
@@ -0,0 +1,14 @@
+development:
+  secret_key_base: b8034ca51af7b188d834a07f0cfc24ea74158c16751604db7394a38d86e14bd549cb18fa843905118a544f92474f700dba55796db2eb26791a0850f2234723de
+  github_oauth_client_id: 1565656fd4d71585be01
+  github_oauth_client_secret: 2862a53781ceec06cd7e561ff0878792fcf20cde
+ 
+test:
+  secret_key_base: 7b70fe9e995b0bc0581c083068fa1159c6227271651ace9194272855c214818bfffc362642ae3af6f325f631bcf688aabc1aca3e06adfe5eb5fd4912a3970e18
+  github_oauth_client_id: 1565656fd4d71585be01
+  github_oauth_client_secret: 2862a53781ceec06cd7e561ff0878792fcf20cde
```
PragTob: huh config/secrets.yml is checked in... it is on gitignore but probably wasn't deleted, I guess we can/should delete it?

spec/features/sign_in_spec.rb
```diff
@@ -36,4 +36,33 @@
       expect(page).to have_content 'Invalid email or password'
     end
   end
+
+  context 'with github' do
+    let!(:person) { create(:person, provider: 'github', uid: '1234567') }
+
+    before(:each) do
+      # https://github.com/intridea/omniauth/wiki/Integration-Testing#omniauthconfigadd_mock
+      OmniAuth.config.test_mode = true
```
PragTob: Wow super cool that you found that, that gives us actual application level testing :+1:

PragTob:
Minor thing about secrets.yml still being there, otherwise looks good to go imo :+1: 

Good job on the tests! :dancer:

config/secrets.yml
berlintam: :scream:

berlintam:
@PragTob I broke it :(

PragTob:
Hehe :) What you really want to do is a "before_script" in the travis.yml that copies over the example secret key base :)

berlintam:
what about that little hack I did?

PragTob:
that is a hack that sort of works, the before_script would be the proper way to deal with this :)

berlintam:
how does one do such a thing?
is putting `"cp secrets.yml.sample secrets.yml"` in the travis.yml enough?

PragTob:
should work :+1:

berlintam:
excite

app/models/person.rb
berlintam: also, if you only have one name, like 'Tam', rorganize it will set your name to 'Tam Tam'. :scream:

PragTob: Which is obviously a feature.

PR #354, opened 2015-07-19
Step 1: use admin boolean instead of role class

berlintam:
I think that [rolify](https://github.com/RolifyCommunity/rolify) is a bit too much for our little app... we only have one role. I initially implemented rolify b/c I was using it at work, but I think that maybe just an `admin` boolean is the way to go. 

This is step one: add the admin boolean & use it
Step 2 will be in a separate PR to remove the role class. I assume it can be merged and deployed one the [rake task](https://github.com/rubycorns/rorganize.it/blob/add-role-boolean-to-people/lib/tasks/set_admin_role.rake) from step one has been run.

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -4,15 +4,14 @@ class AdminsController < ApplicationController
   before_action :ensure_admin_powers
 
   def create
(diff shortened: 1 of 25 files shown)


berlintam:
all is now in the other PR.

PR #355, opened 2015-07-19
Step 2: Remove rolify

berlintam:
this is step2!

One question: should anything be done about this migration?
https://github.com/rubycorns/rorganize.it/blob/master/db/migrate/20140517093641_rolify_create_roles.rb

If so... wat?


berlintam:
oops, I accidentally pushed our secret keys. I've changed them.

app/controllers/application_controller.rb
```diff
@@ -16,6 +16,7 @@ def configure_permitted_parameters
       :picture,
       :working_on,
       :workshop_coach,
+      :admin
```
til: Minor nitpick, but in the code above whoever wrote this was following the one-comma-after-each-line convention. Which totally works in ruby, yay! So I'd say when adding an attribute, it should continue with that convention.

berlintam: haha, ok, thanks Til. I'll add a comma :) (ps - i didn't know/notice that was possible, cool!)

app/views/dashboards/show.html.haml
```diff
@@ -18,10 +18,12 @@
                   = f.hidden_field :admin_id, value: admin.id
                   %button.link.no-btn
                     %span= fa_icon "magic", text: "un-adminify"
+
           - if @non_admins.length > 0
-            = form_tag dashboard_admins_path, method: 'POST' do
+            = form_tag dashboard_admins_path, method: 'create' do
```
til: Are you sure? http://api.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-form_tag

berlintam: nope! not at all!

db/migrate/20150719091734_add_admin_to_people.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdminToPeople < ActiveRecord::Migration
+  def change
+    add_column :people, :admin, :boolean, default: false
+  end
+end
```
til: With this version, can the attribute also be NULL in the database? If yes, what would it mean? Would it make sense to allow NULL?

spec/controllers/posts_controller_spec.rb
```diff
@@ -60,8 +60,7 @@
       context 'as an admin' do
 
         before do
-          person.add_role :admin
-          person.save
+          person.admin = true
```
til: Instead of changing the existing person, those contexts could also simply create a new admin, since we have the nice factory for this already. Somehow like that, instead of the `before` block: `let(:person) { create(:admin) }`.

But it was like this in the code before already, so maybe don't change it in this PR for the sake of no unrelated changes. Or change it, for the sake of always leaving the code a bit better than you found it. Whatever.

berlintam: :+1:

til:
Regarding the migration - well, it adds two tables, and those two tables are useless after these changes, so there should be a new migration that removes them. BUT! Except for the information  in them who is an admin, which first needs to be carried over to the new schema.

The rake task is a good approach, but maybe more suited for optional data changing stuff. In this case, carrying over the admin information is crucial, so maybe putting it into a migration would be a better fit.

Just one thing: in migrations, there should not be code that relies too much on application code, because that can go away, and thus is fragile. So instead of relying on the method `Person.with_role(:admin)`, which comes from rolify, which is going away, the migration should use the contents of the tables in the database directly. You will need to figure out how rolify stores the information there, and replicate that.

It occured to me now that this rake task will probably not work either if I understand correctly, because the rolify gem was removed, so `with_role` is not available anymore.

til:
Some remarks, but in general: great stuff, it's nice to see how much simpler the code becomes with this. Here, have some cake already: :cake:

berlintam:
Thanks Til! Regarding the migration and using `with_role`, yes. I was thinking though, that first we would merge step 1, and do the migration (which uses rolify), and then merge step 2, which removes rolify. I guess thou.... what you say makes more sense and is less fragile. I shall tackle it tomorrow at da cornz.

Thanks for the cake & for the review! nom nom nom!

til:
I was about to write that the PR would then have to be split into two
parts, but then saw that you already did that. Well duh, maybe I should
have read the title more carefully ;-) ! OK, that will work too, and in
some situations it might even be required to do such a two-step deploy,
but here merging it into one with a migration that changes the data,
should be possible and simpler.

berlintam:
oh man til, i just realized you commented all over my closed pr! :laughing: 
I closed this b/c of the secret key issue and I freaked out :P

db/migrate/20150719091734_add_admin_to_people.rb
berlintam: hmmm, i guess it would not make sense to allow NULL. But when would NULL happen if default were false?

app/views/dashboards/show.html.haml
berlintam: this form actually confuses me quite a bit. I wonder if it makes more sense to update a user? B/c this is in fact what we're doing now.... can't we just scrap this whole thing and update the user instead? (for both creating and and deleting admins)

berlintam:
>  merging it into one with a migration that changes the data,
should be possible and simpler.

Do you mean that I should ditch the rake task and instead put the rake task stuff in the migration that creates the admin boolean? (or a separate one?)

I was under the impression from work ppl (althou I'm willing to admit that there are many different thoughts on the matter), that migrations should be used just for changing the database, and rake tasks should be used for changing the data itself. Thoughts? Have we already talked about this? I'm getting deja vu...

til:
Oops damn, I totally overlooked that this PR was already closed. As I
said, I wasn't reading carefully, ...

Regarding the data changes - yes, there are different opinions on this,
and I think the one more generally adopted is the one you mentioned -
migrations only for changing the schema, and separate rake tasks for
data changes. However, in this particular case, I'd go against the
common consensus and put the data changes in the migration. Mainly
because it saves us from requiring a complicated deploy -> rake ->
deploy choreography.

app/views/dashboards/show.html.haml
til: Yes, updating sounds more adequate now. I think the POST to admin came
From me, and my idea was that it _creates_ a role entry. Questionable
even before this change, but really obsolete now.

db/migrate/20150719091734_add_admin_to_people.rb
til: Probably NULL would not happen with the default being false. But by not allowing it, we'd be sure that it will _never ever_ happen. And nobody would ever be tempted to put in a NULL manually, because clearly we said we don't know what it would mean. I think, not allowing NULL unless they are explicitly meaningful is good practice

PR #356, opened 2015-07-19
Remove Rolify & add admin boolean to users ALL IN ONE GO!

berlintam:
this is step2!

One question: should anything be done about this migration?
https://github.com/rubycorns/rorganize.it/blob/master/db/migrate/20140517093641_rolify_create_roles.rb

If so... wat?


PragTob:
As you did, drop the role table then all should be good :D

berlintam:
@PragTob are you psychic!? You commented literally at the exact moment I pushed the commit to drop the tables!

PragTob:
tobi reads minds

berlintam:
so, NEW PLAN!

everything is in one super awesome migration now. So the two PRs aren't needed anymore!
Instead, let's just merge this one and ditch the other one.

til:
Merged!

berlintam:
thanks for the help @til!

station

PR #359, opened 2015-07-25
Fix migration for postgres

berlintam:
@til is this what you meant?

I dropped my local database & tried it out and it didn't complain, so that's good at least!


til:
Looks good, thanks!

If you have time, please try to deploy again.

berlintam:
OMG IT WORKED! WE ARE ALL STILL ADMINZ! WOOOOOO!
happy

til:
Hooray! :rocket: :cake: well done!

PR #364, opened 2015-08-04
make forms & controllers make sense for person

berlintam:
I was at the airport & refactored. 

* people are now updated in the person controller & not in the user registration controller
* forms are changed & split up to reflect this change

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -1,12 +1,36 @@
 class PeopleController < ApplicationController
 
   before_action :authenticate_person!, except: [:index]
+  before_action :set_person, only: [:show, :edit, :update, :destroy]
 
   def index
     @people = Person.order(:first_name).order(:last_name)
   end
 
   def show
+  end
+
+  def edit
+  end
+
+  def update
+    if @person.update_attributes person_params
+      redirect_to person_path(@person), notice: 'You updated your profile!
+      Please endulge in some cake and pat yourself on the back for
+      a job well done.'
+    else
+      render 'edit'
+    end
+  end
+
+  private
+
+  def set_person
     @person = Person.find(params[:id])
+  end
+
+  def person_params
+    params.require(:person).permit(:first_name, :last_name, :email, 
+    :picture, :twitter, :working_on, :workshop_coach, :admin)
   end
 end
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -31,6 +31,6 @@ class PeopleController < ApplicationController
 
   def person_params
     params.require(:person).permit(:first_name, :last_name, :email, 
-    :picture, :twitter, :working_on, :workshop_coach, :admin)
+    :picture, :twitter, :working_on, :workshop_coach)
   end
 end
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -14,12 +14,11 @@ class PeopleController < ApplicationController
   end
 
   def update
-    @person.update_attributes person_params
-    if @person.save
+    if @person.update_attributes person_params
       redirect_to person_path(@person), notice: 'You updated your profile! Please endulge in some cake'\
       'and pat yourself on the back for a job well done.'
     else
-      render 'edit'
+      render action: 'edit'
     end
   end
 
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -14,10 +14,10 @@ class PeopleController < ApplicationController
   end
 
   def update
-    if @person.update_attributes person_params
-      redirect_to person_path(@person), notice: 'You updated your profile!
-      Please endulge in some cake and pat yourself on the back for
-      a job well done.'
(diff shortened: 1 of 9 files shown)


app/controllers/people_controller.rb
```diff
@@ -1,12 +1,36 @@
...
+  end
+
+  def edit
+  end
+
+  def update
+    if @person.update_attributes person_params
+      redirect_to person_path(@person), notice: 'You updated your profile!
+      Please endulge in some cake and pat yourself on the back for
+      a job well done.'
```
til: Great motivational text! Minor style note: this becomes a multiline string with lots of whitespace in it. Maybe use this form instead?

```
'bla'\
'bla'
```
?

til:
O_o ... http://blog.erratasec.com/2012/03/rubygithub-hack-translated.html

app/controllers/people_controller.rb
```diff
@@ -1,12 +1,36 @@
...
+
+  private
+
+  def set_person
     @person = Person.find(params[:id])
   end
+
+  def person_params
+    params.require(:person).permit(:first_name, :last_name, :email, 
+    :picture, :twitter, :working_on, :workshop_coach, :admin)
```
PragTob: this is why @til linked the mass assignment post, you allow the user to set the admin flag for himself :)

berlintam: OR HERSELF TOBZ, OR HERSELF :laughing:

berlintam:
:shipit: ?

app/controllers/people_controller.rb
```diff
@@ -1,12 +1,36 @@
...
 
   def show
+  end
+
+  def edit
+  end
+
+  def update
+    @person.update_attributes person_params
+    if @person.save
```
til: According to http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update , `update` (or the alias of it, `update_attributes`) also validates and tries to save the record. So it should be sufficient to either use `update` or `save`, not both.

Does this work as intended, ie. when there is invalid data, does it correctly redisplay the form, with inline errors?

berlintam: it does now! https://github.com/rubycorns/rorganize.it/commit/0b7df1a5b1f66928ca785843069d15a75783a98a

BUT, I don't understand why. 

To test, I tried to update a person by deleting their name in the hopes that I would get errors. 
When I was using `current_person` in the form, `@person` in the controller wasn't being saved, so I was redirect to the edit page. obviously there were errors here. But errors messages weren't showing up. When i put a binding pry in the error message partial, I could see that the object being passed through, in this case `current_person` did NOT have any errors....

So `current_person` in the form was not being updated (the name wasn't changing and therefore `current_person` was valid), but `@person` from the controller was being rendered on the edit page and was invalid? But... when I looked at `current_person` and `@person`, they were both the same!

Wat?

does this make even a little bit of sense? WHAT IS EVEN A PERSON.

til: I think it does make sense, but it can certainly be confusing because of the way rails does things. But I don't think rails is really to blame for that - maybe this complexity is unavoidable for a framework that wants to support forms with form objects the way rails does. The alternative might be a lot of manual work - coding html forms, validation errors, etc., by hand. That would certainly be more understandable.

The reason why the validation error for the empty name was not showing up was: when POSTing the form with the empty name, the `@person` object is instantiated with those params. As a consequence `@person.save` returns false (and `@person.valid?` would also return false), so the controller runs the branch with `render :edit`, which renders the html form again. However, in `_form.html.haml`, the object being used to fill the values of the form was `current_person`, not `@person`. `current_person` may refer to the same person, but it is still a different instance of the Person class - it was instantiated somewhere by the application controller with `Person.find`, and does not have the modified attributes that were sent in the params hash resulting from the POST.

How is that possible - having two different instances, pointing to the same entity? That's just the way ActiveRecord works - they are different instances of a ruby class, and can be changed independently:

```ruby
> p1 = Person.find(1)
> p2 = Person.find(1)
> p1.first_name
=> "Student"
> p2.first_name
=> "Student"
> p1.first_name = "Blabla"
=> "Blabla"
> p2.first_name
=> "Student"
```

only when the changes are saved to the database and reloaded, will one instance know about them:

```ruby
> p1.save
> p2.reload
> p2.first_name
=> "Blabla"
```

Internally the ActiveRecord instance depends on the value of the `id` attribute, to know which row in its database table to save to or to read from when reloading.

I hope this explains it a bit.

In any case: as form object, one should always use an object that was instantiated in the controller, never something else such as the `current_person` helper method.

berlintam: thanks @til! This makes total sense & clears up a whole bunch of stuff for me. I was totally suspecting that something like this was happening..... but at the same time hoping that such a thing wasn't actually possible.... 

good to know!

:cake:

til: Glad to hear it helped.

berlintam:
:shipit: ?

berlintam:
dance

Issue #298, opened 2015-01-15
group activity status


lislis:
I talked to Sven yesterday and we talked a bit about our little app.
He suggested that we should somehow indicate if a group is still activly meeting.
I think it makes sense and we didn't think about it yet.

Also related to the #282 calendar feature, that maybe we display a warning to an event if the group hasn't been active for a month or so.

The important question is how to measure the activity. Sven suggested some sort of check-in ('Did your group meet this week?'), but I'm not sure if people would use it.

Thoughts?

berlintam:
hm.... interesting....
maybe we can have an active checkbox...
and then maybe an email can be sent out to the group, once a month, and ask them if they're still active and request that their update their status if they're not. 

alternatively/in addition to this, when people sign up for a group, we should somehow make it clear to them that they should contact the group before going to make sure that they're still meeting. Either via an alert of some kind or an email...

PR #365, opened 2015-08-04
adds group inactive tag

sareg0:
@Malwine and I worked on the issue #298 about having an 'group inactive' tag, so that a person knows if the group inactive before joining.

The tests are running, but another pair of eyes, to see if it's missing anything is always good.


berlintam:
:+1: looks good! I guess you'll have to merge master into it thou.

sareg0:
oh gosh... okay, I'll fix that up then push the changes. Probs end of day.

berlintam:
you can do it! (lemmie know if u need help)

sareg0:
@berlintam did I do it correctly? Had to use Tower and Marc, but got it :)

berlintam:
looks pretty good to me! (let's see what Travis says...)

sareg0:
Travis says go!

Gemfile.lock
```diff
@@ -340,6 +340,3 @@ DEPENDENCIES
   turbolinks
   uglifier (>= 1.3.0)
   will_paginate
-
-BUNDLED WITH
-   1.10.5
```
berlintam: my only concern is that this needs to be back in the gemfile.lock.

Brigitte had the same problem on one of her branches (https://github.com/rubycorns/rorganize.it/pull/350), and @til  said:

"Just revert this change to this file, and push that as new commit to this branch?

Bonus points for squashing it, but can be done later.

Oh, and run a gem install bundler locally to make sure you have the most recent one installed. Then you should be safe from getting this change applied automatically again, I think there were a few recent versions of bundler that always changed this section, but the most recent one only does it if the Gemfile.lock also changed elsewhere."

berlintam: I think this stackoverflow comment can help you with that...
http://stackoverflow.com/questions/215718/reset-or-revert-a-specific-file-to-a-specific-revision-using-git

PragTob:
:+1:

Malwine:
giphy

PR #368, opened 2015-08-25
get lots of info for coaches page

berlintam:
This PR adds:

* willing to travel checkbox
* city field
* country field

to user. 

It also explains a bit about what the Rails Girls Coaches feature is, and what it means to click all these boxes. 

Adding city & country will help us/organizers filter the coaches better.

Til and I also discussed a bit what to do about email addresses: we decided not to display the email addresses of users, and instead if a Rails Girls organizer wants to USE our coach list, they have to contact us and prove that they are trustworthy, and then we will send the email addresses to them. I think this is cool for now and can totes be expanded upon in the future if need be.


til:
:+1:

PR #385, opened 2015-09-08
merge user & github accounts via profile

### LINKED ISSUE #382: Allow connect to github when logged in
> As a user that is logged in, I want to connect my account to github so I can use the convenience of the github login for logging into the website.

To get this done, it'd be nice to have a button in the profile/settings that says something like "connect to github" that then executes the github connection. I'm not sure if we can send data to github that they then send back to us so that we know who that github account belongs to - have to figure that out.

Which means set the provider and uid to the appropriate value.

@berlintam

berlintam:
closes #382 like a boss. 

There's now a link on someone's profile page (if they're logged in and if they're not yet merged w/ github) offering a person an opportunity to link their rorganize account with their github account. 

If they click on this button, they will be able to login with github in da future. 

woot. Check it @PragTob :smile:

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -3,15 +3,15 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def github
     # You need to implement the method below in your model (e.g. app/models/user.rb)
     @person = Person.from_omniauth(request.env["omniauth.auth"])
-    rorganize_person = Person.where(email: @person.email).where(provider: nil).first
+    rorganize_person = Person.find_by(email: @person.email, provider: nil)
 
     if rorganize_person.present?
       rorganize_person.merge_with_github!(@person)
       redirect_to rorganize_person
-      set_flash_message(:notice, :success, kind: "Github") if is_navigational_format?
+      set_flash_message(:notice, :success, kind: "GitHub") if is_navigational_format?
     elsif @person.persisted?
       sign_in_and_redirect @person, event: :authentication #this will throw if @user is not activated
-      set_flash_message(:notice, :success, kind: "Github") if is_navigational_format?
+      set_flash_message(:notice, :success, kind: "GitHub") if is_navigational_format?
     else
       session["devise.github_data"] = request.env["omniauth.auth"]
       redirect_to new_person_registration_url
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -3,8 +3,13 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def github
     # You need to implement the method below in your model (e.g. app/models/user.rb)
     @person = Person.from_omniauth(request.env["omniauth.auth"])
+    rorganize_person = Person.where(email: @person.email).where(provider: nil).first
 
-    if @person.persisted?
+    if rorganize_person.present?
+      rorganize_person.merge_with_github!(@person)
+      redirect_to rorganize_person
+      set_flash_message(:notice, :success, kind: "Github") if is_navigational_format?
+    elsif @person.persisted?
       sign_in_and_redirect @person, event: :authentication #this will throw if @user is not activated
       set_flash_message(:notice, :success, kind: "Github") if is_navigational_format?
     else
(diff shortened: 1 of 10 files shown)


berlintam:
wait, don't check it. I forgot to test the new `merge_with_github!` method. :cry:

berlintam:
@PragTob @til this is done & ready! If anyone thinks there's still something that needs to be done here, feel free to take it over while I'm on vacation, or leave it until I get back. Otherwise, pls merge!

berlintam:
it looks like this when your account is not linked w/ github:

screen shot 2015-09-14 at 13 29 45

it looks like this when your account IS linked w/ github:
screen shot 2015-09-14 at 13 29 59

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -3,8 +3,13 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def github
     # You need to implement the method below in your model (e.g. app/models/user.rb)
     @person = Person.from_omniauth(request.env["omniauth.auth"])
+    rorganize_person = Person.where(email: @person.email).where(provider: nil).first
```
PragTob: you could just do a hash with both the email and provider key, or even just that with `find_by` (and no first)

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -3,8 +3,13 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def github
     # You need to implement the method below in your model (e.g. app/models/user.rb)
     @person = Person.from_omniauth(request.env["omniauth.auth"])
```
PragTob: `@person` is only used in the elsif, isn't it? :)

PragTob: doh of course it isn't.... sorry forget what I said. I'm tired.

app/models/person.rb
```diff
@@ -51,6 +51,12 @@ def self.from_omniauth(auth)
     end
   end
 
+  def merge_with_github!(github_person)
+    self.provider = github_person.provider
+    self.uid = github_person.uid
+    self.save
```
til: There is `save` and `save!`. The difference is what happens when a validation fails: the first one just returns false, the second one throws an exception. Did you use `save` on purpose, or would you agree that maybe `save!` would be better here? Generally I think it`s better to use `save!` when not checking for the return value explicitly.

til: Why? Because if there is an error, it will happen earlier and will therefore be easier to find.

spec/features/sign_in_spec.rb
```diff
@@ -65,4 +64,36 @@
...
+          image: ""
+        }
+      )
+    end
+
+    before do
+      sign_in person
+      visit person_path(person)
+      find("#github-button").click
+    end
```
til: `before` is meant to do preparatory work before the actual spec, that is necessary but not directly related to the subject of the spec. The steps in this `before` block are directly related to the topic of the spec, so should be inside the example (`it`), in my opinion.

spec/models/person_spec.rb
```diff
@@ -146,4 +146,16 @@
     end
   end
 
+  describe '#merge_with_github!' do
+    let(:github_person) { build :person, provider: 'github', uid: '123456' }
+    let(:person) { build :second_person }
+
+    before { person.merge_with_github!(github_person)}
```
til: Same remark as above.

til:
GitHub is correctly spelled with an uppercase H: "GitHub". I think at least in the user visible text it should be spelled correctly, otherwise we risk summoning the spanish spelling inquisition.

app/models/person.rb
PragTob: :+1:

spec/features/sign_in_spec.rb
PragTob: agreed and talked about it with tam here :)

berlintam:
ok @til & @PragTob ! moar comments? or is this GitHHHHHub business good to go?

til:
Apart from the minor typo looks good to me. Do you want to rebase and squash? Or should we just merge it?

berlintam:
thanks @til! I say let's just merge :)

til:
Mrrgd! You wanna deploy?

Issue #376, opened 2015-09-01
Extract groups filter into a partial


berlintam:
This way we can hopefully re-use it for the coaches index page!

related to https://github.com/rubycorns/rorganize.it/issues/369

Issue #369, opened 2015-08-25
add filter to coaches page


berlintam:
So that users can be filtered according to city/country.

PragTob:
:+1:

sareg0:
I assigned to myself since I am working on this.

PR #386, opened 2015-09-29
Coach page filter and filter partial

sareg0:
This touches on two issues #376 and #369

I feel like there is something missing, but please do take a look into it.... thoroughly!

Looking forward to the feedback :)

### DIFF
--- a/app/views/groups/index.html.haml
+++ b/app/views/groups/index.html.haml
@@ -34,22 +34,4 @@
       %h1.page-header Hey listen!
       %p If you don't see your project group - feel free to add it!
       = link_to "register new group", new_group_path, class: 'btn btn-primary'
-      
-      %h2.page-header Filter by
-      = form_tag groups_path, method: 'get' do
-        .row
-          %label.col-md-2.control-label city
-          .col-md-9
-            %select{name: :city, class: 'form-control'} 
-              - @group_cities.each do |group|
-                %option{value: group.city}= group.city
-        .row
-          %label.col-md-2.control-label country
-          .col-md-9
-            %select{name: :country, class: 'form-control'} 
-              - @group_countries.each do |group|
-                %option{value: group.country}= group.country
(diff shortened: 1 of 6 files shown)


app/views/groups/index.html.haml
```diff
@@ -34,22 +34,4 @@
...
-          %label.col-md-2.control-label country
-          .col-md-9
-            %select{name: :country, class: 'form-control'} 
-              - @group_countries.each do |group|
-                %option{value: group.country}= group.country
-        %button{type: 'submit', class: 'btn btn-primary'}
-          filter
-
-      
+      =render 'filter'
```
til: Minor code style remark: usually in haml, a `=` should be followed by an empty space: `= render 'filter'`

See also http://haml.info/docs/yardoc/file.REFERENCE.html#inserting_ruby_ (although they don't specifically mention it there, but give examples)

til:
Looks good to me so far.

One thing - the code in both the `_filter.html.haml` partials looks very similar to me. They do have a few important differences, but mostly they are the same haml code. As you know, one of the worst enemies of bug-free, beautiful maintainable code is duplication. So what to do? If you want, I suggest you try to merge the two filter partials into one shared partial. I think it should be possible, but I haven't thought it through fully, so it would be an experiment. See

http://guides.rubyonrails.org/action_view_overview.html#partials

for some hints what can be done with partials.

sareg0:
okay, thanks for the help @til. I'll work on trying to just have on partial over the coming days :) Agreed, on that it is duplicative. Thanks for the tip about the haml syntax too. I'm not super familiar with it, so reading the docs is probz a good idea :)

sareg0:
boom! squashed. Thank you @PragTob

til:
Hm hm, now the commit history has a commit " starts fix of #369 " and "fix of #369" and a merge commit in between ... wouldn't it be possible to squash it all into one? At least if you want to squash, you might squash it all the way.

Issue #384, opened 2015-09-08
Support second names etc.


PragTob:
See [this comment](https://github.com/rubycorns/rorganize.it/pull/352/files#r38303006)

The `name=` method only assumes that the name is of the form "firstName lastName" but there is also "firstName secondName lastName" and stuff like that :)

PR #387, opened 2015-09-29
account for multiple names & only one

berlintam:
closes https://github.com/rubycorns/rorganize.it/issues/384

Now people can have one name, two, three, four, or even 100!
All names except for the last are considered to be first names. 

Suggestions welcome!

### DIFF
--- a/spec/models/person_spec.rb
+++ b/spec/models/person_spec.rb
@@ -81,17 +81,26 @@ describe Person do
 
     it 'does not double the name if there is only one' do
       subject.name = 'Ruby'
+
       expect(subject.name).to eql 'Ruby'
+      expect(subject.first_name).to eql 'Ruby'
+      expect(subject.last_name).to eql nil
     end
 
     it 'allows for three names' do
       subject.name = 'Ruby Fabulous Rubycorn'
+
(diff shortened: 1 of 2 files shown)


spec/models/person_spec.rb
```diff
@@ -73,6 +73,28 @@
...
+      subject.name = 'Ruby Fabulous Rubycorn'
+      expect(subject.name).to eql 'Ruby Fabulous Rubycorn'
+    end
+
+    it 'allows for many many names' do
+      subject.name = 'Ruby Fabulous Rubycorn What Does It Even Mean'
+      expect(subject.name).to eql 'Ruby Fabulous Rubycorn What Does It Even Mean'
+    end
+  end
+
```
til: These specs set the name, and then expect the name to be what it has been set to. While this is a useful thing to check in this case, because behind the scenes crazy elaborated split and join operations are happening, it does not look too useful from just looking at this spec. What about extending the expectations to the values of `first_name` and `last_name` as well, to make sure that the splitting works as desired?

berlintam: sure, makes sense. I also thought as I was writing this test that it seemed a bit weird.

berlintam:
ok @til, done!

berlintam:
thanks @til !

Issue #353, opened 2015-07-12
add website to user


berlintam:
& let the user fill in these field in their profile

Malwine:
I will start working on it.

berlintam:
@Malwine did it!

PR #388, opened 2015-10-06
(WIP) Enable adding website to profile

Malwine:
Enable adding website to profile as described in [#353](https://github.com/rubycorns/rorganize.it/issues/353) 

TODO:
- [x] Make a website partial that takes care of turning the website string into a link

- [x] Make the labels clickable for accessibility reasons

- [x] Parse website for `https://` or `http://` and if not present add `http://`

Additionally:
- we adjusted the readme on how to run the specs
- we changed the label tag of the person/edit to be connected to the input in order to put the focus when clicking the label

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,9 @@ Run the server
 
 et voilá
 
+Run all the tests 
+
+    bundle exec rspec
 
 
 ### Git: branches 
--- a/app/views/people/show.html.haml
+++ b/app/views/people/show.html.haml
@@ -27,6 +27,10 @@
         %p
           Twitter:
           = render partial: "shared/twitter", locals: { object: @person }
+      - if @person.website?
+        %p
+          Website:
+          = render partial: "shared/website", locals: { object: @person }
       - if @person.workshop_coach?
         %p
           I am willing to coach at Rails Girls workshops
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -34,6 +34,6 @@ class PeopleController < ApplicationController
 
   def person_params
     params.require(:person).permit(:first_name, :last_name, :email,
-    :picture, :twitter, :working_on, :workshop_coach, :city, :country, :willing_to_travel)
+    :picture, :twitter, :website, :working_on, :workshop_coach, :city, :country, :willing_to_travel)
   end
 end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -40,6 +40,8 @@ class Person < ActiveRecord::Base
 
   validates :first_name, presence: true
 
+  before_validation :prepend_http
+
   scope :admin, -> { where(admin: true) }
   scope :by_country, -> (country) { where(country: country) }
   scope :by_city, -> (city) { where(city: city) }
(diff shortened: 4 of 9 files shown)


Malwine:
@til I just realized that the problem I had when we left was that I was asking the `edit` page for the website instead of the `show`. I didn't realize that I would go to `show` after save. So I added it there and the test passed. Yay. :100:

README.md
```diff
@@ -63,6 +63,9 @@ Run the server
 
 et voilá
 
+Run the test
```
til: Maybe even say _all tests_? I mean it's more than one ...

til: Anyway, thanks for adding it!

app/views/people/show.html.haml
```diff
@@ -27,6 +27,10 @@
         %p
           Twitter:
           = render partial: "shared/twitter", locals: { object: @person }
+      - if @person.website?
+        %p
+          Website:
+          #{@person.website}
```
til: There are different ways in haml to run ruby code and insert it into the page. This is one of them, and works, but is mostly useful when you want to embed something within a lot of static text. In this case, when it's just on one line, you might as well use `= @person.website`.

til:
Cool stuff, happy to hear the spec works now.

Debugging these things can be hard, because it's like operating a
browser without seeing anything. You need to insert debugging output at
the right places to find out what's happening. E.g.

puts page.body

is a very simple one. There is also

save_and_open_page

in capybara (which may or may not work).

and

puts current_path

to see where the browser currently is.

Regarding turning the string into a link - as usual there are many
possibilities how to do that, what do you think about using "link_to"?

app/views/people/show.html.haml
Malwine: Ok, but I will maybe make a partial like the twitter partial. What do you think? Because we want it to become a link, right? :blush:

Malwine:
Ah now I saw this comment. Sure, we can use link to.

Malwine:
Travis is not happy :smile:

README.md
Malwine: I will adjust it then :smile:

app/views/people/show.html.haml
til: Ah yes, you said that before. A partial is cool for this, yes, go ahead!

Malwine:
@til Making the labels clickable working with the exception of the `working on` area. I am wondering why that happens...

til:
Cool. Regarding working on, a hint: it works in html through connecting
the "for" attribute of label with the "id" attribute of the input field.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

I'm sure you can fix this as well ...

Malwine:
I need to figure out how to proceed with the website. Want to do that the next days. We need to check wheather the webite string has http:// or not, right? Otherwise the link won't work. So I was wondering if we should only store the domain without www, http:// or https:// and add it when creating the link? Similar to the twitter handle procedure.

til:
I would advise against cutting off https or even www - some links might only work this way, others only that way, modifying the link has the risk of breaking it.

Maybe it is sufficient to give a little hint in the form, that the link needs to be complete, ie. including http or https?

til:
Here are the links to the documentation about activerecord validations, which I mentioned yesterday:

http://api.rubyonrails.org/classes/ActiveModel/Validations/Callbacks/ClassMethods.html#method-i-before_validation

http://guides.rubyonrails.org/active_record_validations.html

If you have any questions, just ask!

Malwine:
Hi @til 
Thanks for the links.
I just commited the before validation parsing. Tests are passing. At least locally :smiley: 
Would be cool if you can have a look. 
Maybe there is a way to make the if statements nice?

til:
:+1:

Issue #389, opened 2015-10-06
Issue for Malwine - Prettify default pictures


Malwine:
I gonna do it! :100: 

(Just some minor changes)

PR #390, opened 2015-10-07
Prettify the default avatar and group pictures

Malwine:
The default pictures needed some rework.
Closes https://github.com/rubycorns/rorganize.it/issues/389

Preview:

# People index
Before:
screen shot 2015-10-07 at 21 45 42

After:
screen shot 2015-10-07 at 21 25 17

# Groups index
Before:
screen shot 2015-10-07 at 21 46 49

After:
screen shot 2015-10-07 at 21 37 57

# Nav bar
Before:
screen shot 2015-10-07 at 21 48 22

After:
screen shot 2015-10-07 at 21 45 05


til:
Fantastic! So pretty, much wow!

Lots of :cake: for you, @Malwine

til:
So, who wants to deploy?

Btw, I think those images should be in `app/assets/images`, not `public`. The way it is now, it might be that some browsers will cache the old versions after the deploy and still show those.

berlintam:
such wow!

Issue #271, opened 2014-10-08
prevent scroll-zooming on the map (super annoying)


lislis:
put ```scrollWheelZoom: false``` to map options

sareg0:
okay peoples. I have researched this, and I cannot figure out how to disable the scroll zoom using the leaflet gem.

I am of the opinion we leave this be. I know it's annoying if a map is taking up the whole page and a really big section, and you're stuck in it and can't get down to the next page section or up because you can't get out of the map, or if you are on a mobile and can't get past map because your fingers just scroll the map. gaaahhhhhhhhhhh!!!!

Okay, so I wanna fix the map to not be scrollable on smaller screens unless you pinch. I will figure it out, but I could just do with some help is what I'm saying in a round about sort of a way.

HAAALLLP!!

sareg0:
Okay, so I agree that having scroll zoom on a map is super annoying, particularly when the maps takes up the whole view port, and you seem to be 'stuck' in it.

However I think that the tile is small enough on the RoRganize.it app, that one will be ok on Desktop, but mobile is particularly problematic; the map does indeed take up the whole view on mobile and if you reach down with your thumb and try to scroll the page up it does it go scrolling off into nowhere.

So for the use case on mobile at the very least I was interested in implementing this.

However, @lislis when I add `scrollWheelZoom: false` to the map options in the 'groups_helper' nothing happens. Am I putting it in the wrong place? 

I faffed about with it for quite some time last night, and couldn't get it to stop scroll zooming.

sareg0:
So after speaking with @lislis and looking at the Gem, we are of the opinion that the scrollWheelZoom property simply isn't in the leaflet-rails gem. Apart from that some pull requests and issues have been sitting for quite some time.

This makes us hesitant to make a pull request.

With this in mind, we thought it might be better to simply just do the map ourselves using JavaScript. 

This would involve implementing the library into the project as a part of the assets of the project, and removing the gem.

We might do this between now and the next Rubycorns, or at the Rubycorns on Tuesday.

sareg0:
I'm closing this puppy up now.
PR #406 has prevented the scrolling with a mouse, and has also replaced the leaflet gem for ruby, with the CDN for the library, and a JavaScript.
BOOM!

til:
Yeah, congrats! :smile:

PR #406, opened 2015-11-17
starts fix of #271

sareg0:
This is just a start to our effort to move away from the leaflet rails gem. It doesn't ACTUALLY  solve the original problem of the scroll zoom, but I wanted to present it, to make sure I was heading in the right direction

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,9 +60,6 @@ gem 'jbuilder', '~> 1.2'
 # Some Modernizr goodness
 gem "modernizr-rails", "~> 2.6.2.3"
 
-#use osm maps
-gem 'leaflet-rails', '~> 0.7'
-
 # get coordinates from adress
 gem 'geocoder'
 
--- a/app/views/groups/show.html.haml
+++ b/app/views/groups/show.html.haml
@@ -95,8 +95,7 @@
     .col-md-4
       #group-map
         - if @group.address?
-          .map
-            = groups_map @group
+          #map{data: {latitude: @group.latitude, longitude: @group.longitude}}
         - else
           = image_tag('follow_your_dreams.gif')
 
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -5,8 +5,10 @@
   <title>RoRganize.it | Find a Rails Girls Study Group</title>
   <%= stylesheet_link_tag    "application", media: "all", "data-turbolinks-track" => true %>
   <%= csrf_meta_tags %>
-  <%= javascript_include_tag "application", "data-turbolinks-track" => true %>
   <link href='http://fonts.googleapis.com/css?family=Pacifico' rel='stylesheet' type='text/css'>
+  <link rel="stylesheet" href="//cdn.leafletjs.com/leaflet/v0.7.7/leaflet.css" />
+  <%= javascript_include_tag "application", "data-turbolinks-track" => true %>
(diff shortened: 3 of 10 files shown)


til:
Looks good to me, this is heading in the right direction! Go go go, solve the scroll zoom problem! :grin: 

One thing: we might switch to https sooner or later, so in order to be prepared for that, you could try to change all external references to be protocol relative, so e.g. `//{s}.tile.openstreetmap.se/hydda/full/{z}/{x}/{y}.png` instead of `http://{s}.tile.openstreetmap.se/hydda/full/{z}/{x}/{y}.png`. I think it should work.

Gemfile
```diff
@@ -60,8 +60,8 @@ gem 'jbuilder', '~> 1.2'
 # Some Modernizr goodness
 gem "modernizr-rails", "~> 2.6.2.3"
 
-#use osm maps
-gem 'leaflet-rails', '~> 0.7'
+# #use osm maps
+# gem 'leaflet-rails', '~> 0.7'
```
til: You might as well delete the two lines right away, since this is what this branch is about: deleting the dependency on the leaflet gem. No big deal, but the deletion is even more obvious in the diff than the commented out change, and having less lines around that need to be changed before a merge, the better.

app/views/groups/show.html.haml
```diff
@@ -95,8 +95,7 @@
     .col-md-4
       #group-map
         - if @group.address?
-          .map
-            = groups_map @group
+          #map{ data: {latitude: @group.latitude, longitude: @group.longitude}}
```
til: Most minor nitpick: here is either a space too much or too little ...

sareg0: I think I have fixed this in the changes I have pushed... Just let me know if I corrected the right thing.

sareg0:
ok @til I have turned off the scrollwheelzoom. 

If you could just take one last gander, that would be grand... I am scared of squashing from my last experience, but I will try to figure out how to do it again, if you would so like :)

app/views/groups/show.html.haml
til: Yup, looks good now.

app/views/layouts/application.html.erb
```diff
@@ -5,8 +5,10 @@
   <title>RoRganize.it | Find a Rails Girls Study Group</title>
   <%= stylesheet_link_tag    "application", media: "all", "data-turbolinks-track" => true %>
   <%= csrf_meta_tags %>
-  <%= javascript_include_tag "application", "data-turbolinks-track" => true %>
   <link href='http://fonts.googleapis.com/css?family=Pacifico' rel='stylesheet' type='text/css'>
+  <link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.css" />
+  <%= javascript_include_tag "application", "data-turbolinks-track" => true %>
+  <script src="http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.js"></script>
```
til: Maybe just use `//` instead of `http://` in these two cases too?

til: An alternative would be copying these two files into our app and serve it ourselves directly.

til:
Squash if YOU want to squash - I'm fine with it either way :smile:

Reminder: something along the lines of `git rebase -i master`, then choosing `squash` instead of `pick`, then saving the file.

Malwine:
:boom: :100:

PragTob:
Safe to merge once the tests pass :+1:

Issue #232, opened 2014-07-25
blog next, previous buttons


berlintam:
It would be cool to have `next` and `previous` buttons on each posts#show page so that our eager readers can easily manoeuvre from one blogpost to the next.

mariemariemarie:
we discovered, there is already pagination in the code in the post controller.
It is currently 20 posts/page, we are going to reduce this to 5.
currently there are no tests for this, we will write one.

berlintam:
I was thinking about this again since last night, and I think this issue was more meant as having next, previous buttons on the posts show pages. 

A post show page is this: http://rorganize.it/posts/we-heart-emojis
As opposed to the index page, which is this http://rorganize.it/posts

So someone can click on an individual post, and then hit next, and go to the next one, or back, and go back one. 

I think that the pagination that's already there is just for the index page.

Anyway, sorry for the confusion! and yay tests :)

sareg0:
There has been some complications with this, and I need to do two thing in order to finish fixing

- I am considering turning the published_on column into a datetime object. Currently we are sorting by created_at. This doesn't work, as the blog post may have been created_at a certain time and stored as a draft for sixmonths... So it messes up the order. I want to order by published_on, but if there are multiple posts published on that date, then we need to sort by time of the publishing. I am not sure about the implications of this
- I have discovered that the blog is mistaking 'drafts' as something that should be listed, at least locally. It doesn't seem to be that case in production Because of the 'show_date' method in the helpers. It's using saying that if there is no published on date one should show 'draft created at  on'... It's a bit misleading, as if there is no published on date, you don't want to see that post at all. I know it's only locally, but I figure it's something worth solving. 
Proposal: remove the draft column. Identify drafts by the fact that they don't have a 'published_on' date in the table to indicate that they are a draft.
These are musings from my discoveries today. Will try to act on them

til:
Makes sense - I suggest to then rename the column to published_at, to
follow the convention "_on" for dates, "_at" for datetimes.

berlintam:
#yolo merged

PR #407, opened 2015-12-01
Adds 5 post per page pagination + test

sareg0:
Hey, this is the start of a fix for #232 

Taking down the number of blog posts per page to 5, and writing a test.

Since I am still shakey with tests, I wanted to submit this first, and then implement the next improvement from that issue; adding a back and next button to each Post#show

Let me know if I have done this right, first.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -51,8 +51,6 @@ ActiveRecord::Schema.define(version: 20151013180304) do
     t.boolean  "inactive",           default: false
   end
 
-  add_index "groups", ["slug"], name: "index_groups_on_slug", unique: true
-
   create_table "memberships", force: true do |t|
     t.integer  "group_id"
     t.integer  "person_id"
@@ -86,11 +84,11 @@ ActiveRecord::Schema.define(version: 20151013180304) do
     t.text     "working_on"
     t.boolean  "workshop_coach"
     t.boolean  "admin",                  default: false, null: false
-    t.string   "provider"
-    t.string   "uid"
     t.boolean  "willing_to_travel",      default: false
     t.string   "city"
     t.string   "country"
+    t.string   "provider"
+    t.string   "uid"
     t.string   "website"
   end
 
@@ -109,8 +107,6 @@ ActiveRecord::Schema.define(version: 20151013180304) do
     t.string   "slug"
     t.date     "published_on"
   end
-
-  add_index "posts", ["slug"], name: "index_posts_on_slug", unique: true
 
   create_table "topics", force: true do |t|
     t.text     "body"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -51,6 +51,8 @@ ActiveRecord::Schema.define(version: 20151013180304) do
     t.boolean  "inactive",           default: false
   end
 
+  add_index "groups", ["slug"], name: "index_groups_on_slug", unique: true
+
   create_table "memberships", force: true do |t|
     t.integer  "group_id"
     t.integer  "person_id"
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -51,8 +51,6 @@
     t.boolean  "inactive",           default: false
   end
 
-  add_index "groups", ["slug"], name: "index_groups_on_slug", unique: true
-
```
til: Careful with the dreaded schema.rb file. It changes, when you least expect it, and sneaks behind your back into a git commit. :ghost: :scream: 

These changes are unrelated to your commit. The `add_index` statements should not be removed. The best way to avoid this in the future, is to selectively add each file to git on its own before comitting, with `git add <filename>` calls, instead of just `git add -a`. I know it can be tedious in the beginning, but after some time it even becomes fun! At least I find it fun - it's a last clean up, review of what I did, remove debug output etc., and prevent the sneaky schema.rb from getting committed.

The best way to remove this from this PR: I suggest to add a new commit that reverts them, and later before the PR gets merged, they can be squashed.

db/schema.rb
```diff
@@ -86,11 +84,11 @@
     t.text     "working_on"
     t.boolean  "workshop_coach"
     t.boolean  "admin",                  default: false, null: false
-    t.string   "provider"
-    t.string   "uid"
     t.boolean  "willing_to_travel",      default: false
     t.string   "city"
     t.string   "country"
+    t.string   "provider"
+    t.string   "uid"
```
til: Here only the order of some lines changed, nothing was actually removed. This is somewhat random, depending on the mood of your local database. In any way, it doesn't make sense to be included here.

spec/controllers/posts_controller_spec.rb
```diff
@@ -30,6 +30,17 @@
...
+      before :each do
+        7.times { Post.create title: 'random', description: 'important stuff', draft: false }
+        get :index
+      end
+
+      it 'has 5 published posts' do
+        expect(assigns(:published_posts).to_a.size).to eq 5
+      end
+    end
+
```
til: :clap: that looks great! It creates 7 items, and makes sure that only 5 are assigned, makes sense. In the spirit of the most minimal example, even 6 items would be sufficient I guess.

Also, did you consider creating the test data with a factory method? We are using FactoryGirl for that ( http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md ), which saves you from having to specify mandatory attribute values that are not relevant to a test. E.g. `create(:post)` instead of `Post.create ...`.

spec/controllers/posts_controller_spec.rb
```diff
@@ -30,6 +30,17 @@
...
     end
 
+    context 'with pagination' do
+      before :each do
+        7.times { Post.create title: 'random', description: 'important stuff', draft: false }
+        get :index
+      end
+
+      it 'has 5 published posts' do
+        expect(assigns(:published_posts).to_a.size).to eq 5
```
til: Is `to_a` necessary?

spec/controllers/posts_controller_spec.rb
berlintam: WOOOooOOOooT! awesome test @sareg0 !!!!

And yup, factories are awesome! There should already be a post factory lying around somewhere.... :)

sareg0:
Bleh, I had a lot of trouble trying to pull out the shema file from my commits, and I have some more stuff to commit, so I thought I would submit a fresh pull request. Hence, why I close it.

PR #410, opened 2016-01-12
refactor & remove empty strings from dropdown

### LINKED ISSUE #399: remove empty strings from drop down options
screen shot 2015-11-03 at 21 44 32

I guess some ppl have entered empty strings for their cities & countries. These should be deleted in the following methods:
* https://github.com/rubycorns/rorganize.it/blob/master/app/models/person.rb#L57
* https://github.com/rubycorns/rorganize.it/blob/master/app/models/person.rb#L61
* https://github.com/rubycorns/rorganize.it/blob/master/app/models/group.rb#L59
* https://github.com/rubycorns/rorganize.it/blob/master/app/models/group.rb#L63

berlintam:
closes #399

### DIFF
--- a/app/models/concerns/location_filter.rb
+++ b/app/models/concerns/location_filter.rb
@@ -1,0 +1,12 @@
+module LocationFilter
+  extend ActiveSupport::Concern
+
+  included do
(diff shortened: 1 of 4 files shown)


til:
:+1:

PR #411, opened 2016-01-12
add reset button to filter

berlintam:
This PR adds a reset button to the person & group forms (via the filter partial). Resetting in this case just means redirecting back to either the groups page or the people page. 

I'm a bit confused thou about why I had to do `locals[:path]` [here](https://github.com/rubycorns/rorganize.it/compare/add-reset-button?expand=1#diff-133e1f0b65a4cb828384db8ecd25dc41R22) instead of just `path` since I'm passing it through the locals. I thought that maybe it wasn't working because `path` is a reserved word, so I changed it to a bunch of other things trying to fix it, but it didn't work either. Any thoughts are very welcome!

screen shot 2016-01-12 at 20 42 18

### DIFF
--- a/app/assets/stylesheets/custom/_groups_people.scss
+++ b/app/assets/stylesheets/custom/_groups_people.scss
@@ -66,3 +66,8 @@
 .button-class {
   margin-top: 110px;
 }
+
+.reset-button {
+  border: solid 1px #B12200;
+  border-radius: 5px;
+}
--- a/app/views/coaches/index.html.haml
+++ b/app/views/coaches/index.html.haml
@@ -35,4 +35,4 @@
                   %p #{coach.city} #{coach.country}
 
     %aside.sidebar.col-md-4
-      = render 'shared/filter'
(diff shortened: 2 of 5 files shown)


til:
Great that you are taking care of the filter behaviour. As I said in a
previous comment, it's a bit weird right now - and I only saw this PR
later. Having the ability to reset the filter is a good step forward.

You know what would be even cooler? If the reset button only appears
when a filter is active.

Regarding the locals issue, maybe there is a difference between `render
'something', locals: ...` and `render partial: 'something', locals:
...`. I don't know what and why exactly, but remember vaguely that using
`partial: ` somehow works better.

berlintam:
OoOOoOh cooler indeed! I'll shall implement such a thing.

berlintam:
ok @til! Done & squashed :smile:

til:
Looks good, Travis says it works, so let's merge it.

berlintam:
beyonce

PR #412, opened 2016-01-12
remove railsgirlsapp

### LINKED ISSUE #404: Registration email sends to old URL
Oh noes! Thanks @eljojo for discovering this: apparently the link in the email still points to railgsirlsapp.shellycloud.com. Which can be displayed in the browser, but first, it's old and ugly, and second, connecting with GitHub doesn't work from there.

Maybe just grep for railsgirlsapp and find all occurences to see if there are more errors like these.

berlintam:
closes #404 

Unfortunately I couldn't find the link in the email that still points to railgsirlsapp.shellycloud.com. Everything in the email seems fine unless I'm missing something....

### DIFF
--- a/app/views/pages/about.html.haml
+++ b/app/views/pages/about.html.haml
@@ -17,7 +17,7 @@
 
     %p
       The code is open source. Check it out and contribute on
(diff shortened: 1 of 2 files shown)


til:
:+1:

PR #417, opened 2016-01-18
Isolate tests from network connection

### LINKED ISSUE #361: Geocoding error?
This happens when running tests

`Google Geocoding API error: over query limit.`

why? mystery. maybe google will know.

roschaefer:
fixes #361 
Whenever you ```create(:group)``` in the tests you trigger the geocode api in an [after_validation](https://github.com/rubycorns/rorganize.it/blob/master/app/models/group.rb#L50) hook. I decided to use an explicit cassette in every rspec context that uses this group factory. We could also do something similar to [*this*](http://stackoverflow.com/q/16533980). But it adds extra complexity and you get confused if you write another rspec test that triggers a different http request. In my opinion, the ideal solution would be to follow [this suggestion](http://stackoverflow.com/a/7861809).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -97,6 +97,8 @@ group :development, :test do
   gem 'minitest'
   gem 'rspec-its', '~> 1.0.1'
   gem 'timecop'
+  gem 'vcr'
+  gem 'webmock'
 end
 
 # Use ActiveModel has_secure_password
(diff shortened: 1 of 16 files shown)


lislis:
Hey @roschaefer! Thanks again for the PR! We're looking into it right now (lots of things to learn from it and read up on). We'll get back to you shortly :)

lislis:
Awesome!  
aaaaaah

til:
This is cool, thanks!

PR #423, opened 2016-01-26
fixed seeds.rb, missing normalization

veracl:
With a fresh rails installation and a seeded database, when I try to edit a group as a coach and save without changes I get validation errors.

1. log in as coach
2. go to groups index
3. join group as a coach
4. edit group
5. click on update without changes


berlintam:
thanks!!!

berlintam:
picard

PR #424, opened 2016-01-26
WIP Add documentation for database

Malwine:
This PR adds documentation for database and out the deploy instructions for Shelly Cloud.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Make sure you have ImageMagick installed.
 In Terminal run (OS&nbsp;X):
 
     brew install imagemagick
-    
+
 Or for Linux run (or with the package manager of your choice):
 
     sudo apt-get install imagemagick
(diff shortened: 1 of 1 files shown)


Malwine:
@BriocheBerlin und @til

til:
Thanks. I suggest to either not merge the PR open until uberspace is
finally working, or add a note below "Connect to database" that this is
WIP and refers to the new server.

Also, the current state is that the step "Type the password. Ask
@BriocheBerlin for it." is not necessary, because we are connecting as
user ror who is a superuser and can connect without password. But that
might change again later.

Malwine:
@til I agree with everything. I am sure we need to rewrite the whole PR. I actually just opened it so we don't forget to document the process. Once we are ready setting everything up I will adjust the PR (e.g. next week.)

PR #425, opened 2016-01-26
display country name everywhere not iso_code

### LINKED ISSUE #419: make country dropdown list make sense
At the moment we're just using country codes:

screen shot 2016-01-19 at 16 33 22

I dunno about all of you, but I haven't memorized my country codes yet. Maybe I will one day... but it's pretty far down on the priority list tbh.

berlintam:
closes #419

I didn't only change them in the filter dropdown but also in various other places like the person#show and the people indexes etc.... so in THEORY, there shouldn't be any iso codes for countries anymore, it should all just be full country names.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,6 +1,5 @@
 module ApplicationHelper
 
-
   def flash_class(level)
     case level
       when :notice then "info"
@@ -11,7 +10,6 @@ module ApplicationHelper
   end
 
   def markdown(text)
-
     emoji = Redcarpet::Markdown.new(
       MdEmoji::Render.new(filter_html: true),
       no_intra_emphasis: true,
@@ -21,6 +19,10 @@ module ApplicationHelper
     )
 
     raw emoji.render(text)
+  end
+
+  def country_names_for_select(countries)
+    countries.map { |c| [ISO3166::Country[c].name, c] }
   end
 
 end
--- a/app/helpers/groups_helper.rb
+++ b/app/helpers/groups_helper.rb
@@ -17,8 +17,8 @@ module GroupsHelper
     end
   end
 
-  def group_location(group) 
-    [group.street, group.zip, group.city, group.country].select {|x| x.present? }.join(', ')
+  def group_location(group)
+    [group.street, group.zip, group.city, group.country_name].select {|x| x.present? }.join(', ')
   end
 
 end
--- a/app/models/concerns/location_filter.rb
+++ b/app/models/concerns/location_filter.rb
@@ -9,4 +9,8 @@ module LocationFilter
     scope :by_country, -> (country) { where(country: country) }
   end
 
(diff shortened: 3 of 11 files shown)


roschaefer:
These changes cover all of #423, you fixed the same problem :smiley:  The normalization of country codes is done via the [country-select gem](https://github.com/stefanpenner/country_select) in the [people form](https://github.com/rubycorns/rorganize.it/blob/master/app/views/people/_form.html.haml#L56) and the [groups form](https://github.com/rubycorns/rorganize.it/blob/master/app/views/groups/_form.html.haml#L47). Do you think it's useful if we write a validator to check for missing normalization in the group/person model respectively?

berlintam:
hi @roschaefer yup you're right, the same problem was fixed, but that's ok. I merged the PR anyway & there are no merge conflicts, so it's just a happy accident :)

I'm not sure I understand your suggestion. Users can't type in their own countries, so I don't know where a validator would go/how that would work... can you explain a bit more? Thanks!

roschaefer:
Exactly, users can't type in their own countries. That's why I'm not sure if a validator would be a little over-engineered. 

If I think of web applications as a [presentation <-> application <-> database stack] (https://en.wikipedia.org/wiki/Multitier_architecture) apparently the country-select gem catches invalid data at the presentation layer. My gut feeling tells me it should happen at the application layer in the model. So no matter which route the data went (seed data, rails console, another web interface,  some hacker, or us if we create a new form and accidently use a text box), invalid data will be rejected.

Nevertheless, it is out of scope of your pull request :wink:  if we have too much time it' s something we could implement in our ruby workshop on Tuesday. Unless you think "ahh, we don't need that". :smile:

berlintam:
ok, I see what you mean. 

I mean hey, this is a project for learning, so if you wanna take a crack at it, by all means! I am not opposed :)

Thanks for the input & all the PRs!

app/helpers/application_helper.rb
```diff
@@ -23,4 +21,8 @@ def markdown(text)
     raw emoji.render(text)
   end
 
+  def country_names_for_select(countries)
```
berlintam: @til @PragTob I wasn't sure if the application helper was the right place for this.... thoughts?

til: All the helper methods end up in a huge global pile of methods in the
view context anyway, so it doesn't really matter. If you want, you can
add some structure by moving it to a dedicated CountryHelper, but with
only one method I don't think it is important.

PragTob: Til is right (as usual :)). I'd have a slight preference for an own helper related to Comments, Countries or whatever ß just because I like to have dedicated places for functionality. But in this case it's really not important, plus when the application grows we might find a better place to put it.

PragTob:
Looks good to go, also agreed that some sort of validation might be nice but impact is also limited :)

PR #426, opened 2016-01-31
Catch http requests only for test environment

roschaefer:
In development environment ```bin/rake db:setup``` fails because it's surpressing http requests.


PragTob:
:+1: looks good, thank you!

PR #427, opened 2016-02-02
WIP: Trying to add Capistrano

BriocheBerlin:
Tessi suggested to follow the Readme: https://github.com/tessi/capistrano-uberspace
Also said to consider https://github.com/teikei/teikei/blob/master/config/environments/production.rb and Capfile as examples.

Note: Doesn't work yet, some problem with stage deploy.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,8 @@ end
 group :production do
   gem 'pg'
   gem 'shelly-dependencies'
+  gem 'capistrano', '~> 3.4.0'
+  gem 'capistrano-uberspace', github: 'tessi/capistrano-uberspace'
 end
 
 gem 'carrierwave'
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -82,4 +82,19 @@ RailsGirlsApp::Application.configure do
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
+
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -23,6 +23,8 @@ end
 group :production do
   gem 'pg'
   gem 'shelly-dependencies'
+  gem 'capistrano', '~> 3.4.0'
+  gem 'capistrano-uberspace', github: 'tessi/capistrano-uberspace'
```
PragTob: this might already be the problem, the production gems are installed  only on the production system. You need capistrano in the development group because you invoke it from your computer to do something with the production system (e.g. updating it). But the production system doesn't need it :)

config/environments/production.rb
```diff
@@ -82,4 +82,19 @@
...
+         user: 'ror',
+         roles: [:app, :web, :cron, :db],
+         primary: true,
+         ssh_options: {
+           keys: %w{~/.ssh/id_rsa},
+           forward_agent: true,
+           auth_methods: %w(publickey)
+         }
+
+  set :user, 'ror'
```
PragTob: is there a user like that on our production system?

til: Yes, "ror" is the shell user on the uberspace server.

lislis:
Closing this for now. We have a manual deployment written down in the Readme

PragTob:
:'( :'( :'(

Issue #418, opened 2016-01-18
order coaches list better 


sareg0:
Currently the coaches list doesn't really have any rhyme or reason to it! It would be nice to order it by something like alphabetical order.

BriocheBerlin:
@til or @PragTob I made this one very unspectacular change to the coaches controller: 
`@coaches = Person.where(workshop_coach: true).order(:first_name, :last_name)`
It does work in general, only this one name "Zlatan" is not doing what it is supposed to!
<img width="329" alt="screen shot 2016-01-19 at 21 50 43" src="https://cloud.githubusercontent.com/assets/6567555/12431822/f368f7c2-bef7-11e5-9e9c-c47efe3cccc6.png">

berlintam:
best bug ever! I love it!

sareg0:
hehhe, oh Zlatan, always being the trickster :)

til:
If Zlatan would be more humble and write himself as "zlatan", then in your example it would work as expected.

A couple of options:
- it's fine like this for now
- order by lower() in the database
- order by downcase in ruby

Since we are loading all the coaches at once anyway because there not thousands of them, I'd tend towards the last option.

BriocheBerlin:
closed by #428

PR #428, opened 2016-02-09
orders coaches list

BriocheBerlin:
see #418

@PragTob mentioned that a test would be nice...


app/controllers/coaches_controller.rb
```diff
@@ -1,6 +1,7 @@
 class CoachesController < ApplicationController
   def index
-    @coaches = Person.where(workshop_coach: true)
+    # @coaches = Person.where(workshop_coach: true).order(:first_name, :last_name)
```
PragTob: #no_commented code

app/controllers/coaches_controller.rb
```diff
@@ -9,3 +10,6 @@ def index
     @countries = @coaches.countries
   end
 end
+
+
+# user = User.ci_find('username', params[:username])
```
PragTob: #no_commented code

app/controllers/coaches_controller.rb
```diff
@@ -1,6 +1,7 @@
 class CoachesController < ApplicationController
   def index
-    @coaches = Person.where(workshop_coach: true)
+    # @coaches = Person.where(workshop_coach: true).order(:first_name, :last_name)
+    @coaches = Person.where(workshop_coach: true).order_by_name
```
PragTob: we could also implement `workshop_coach` as a scope :) (probably just coach or so)

PragTob:
Yep would love to see a test that showcases basic sorting, the case insesitiveness and the sorting by first and last name :)

app/controllers/coaches_controller.rb
berlintam: OOOH! what is ci_find??

spec/models/person_spec.rb
```diff
@@ -143,6 +143,45 @@
...
+    let!(:workshop_coach) { create :workshop_coach }
+    let!(:second_workshop_coach)  { create :workshop_coach, first_name: 'Zlatan', last_name: 'Zukanovic'  }
+    let!(:third_workshop_coach)  { create :workshop_coach, first_name: 'lolo', last_name: 'lowercase' }
+    context 'combine scopes' do
+      it 'lists in alphabetical order when lowercased name is present' do
+        expect((subject.order_by_name).map(&:name)).to eq ['Corny Chocolate-Bar', 'lolo lowercase', 'Zlatan Zukanovic']
+      end
+    end
+  end
+
```
til: Looks great! But maybe even a bit too much? Spec'ing the combination of two scopes is rather unusual I'd say. That the two scopes together work as expected when combined is the job of rails, and we should trust it that it gets it right.

spec/models/person_spec.rb
```diff
@@ -143,6 +143,45 @@
...
+    let!(:second_person)  { create :second_person }
+
+    it 'lists the correct name' do
+      expect(subject).to contain_exactly workshop_coach
+    end
+
+    context 'for several workshop coaches' do
+      let!(:second_workshop_coach)  { create :workshop_coach, first_name: 'Zlatan', last_name: 'Zukanovic'  }
+      it 'lists in alphabetical order' do
+        expect(subject.map(&:name)).to eq ['Corny Chocolate-Bar', 'Zlatan Zukanovic']
```
til: Are you sure that this is testing the right thing - isn't the order just random?

spec/models/person_spec.rb
```diff
@@ -143,6 +143,45 @@
...
   end
 
+  describe '.workshop_coach' do
+    subject { described_class.workshop_coach }
+    let!(:workshop_coach) { create :workshop_coach }
+    let!(:second_person)  { create :second_person }
+
+    it 'lists the correct name' do
+      expect(subject).to contain_exactly workshop_coach
+    end
```
til: Spec text and what it actually tests are a bit out of sync.

BriocheBerlin:
Can we plühse merge this? 🐶 👀

.ruby-version
```diff
@@ -1 +1 @@
-2.3.0
+2.1.5
```
til: mep

til:
Totally! Except the change to the `.ruby-version` file ...

til:
Tadaaa!

Issue #434, opened 2016-03-05
Inactive groups should not show sign up buttons


lislis:
Current status
<img width="658" alt="screen shot 2016-03-05 at 14 11 00" src="https://cloud.githubusercontent.com/assets/1006478/13547662/1d3960d4-e2dc-11e5-8666-cf23e8bbaa8b.png">

PR #439, opened 2016-03-15
remove signup buttons for inactive groups

berlintam:
closes https://github.com/rubycorns/rorganize.it/issues/434

It is also the end of an era.... bye bye `even_a_button?`. It's been a blast. But unfortunately, you make no sense anymore. 

What are you even doing? I no longer know. 

sad

### DIFF
--- a/app/helpers/groups_helper.rb
+++ b/app/helpers/groups_helper.rb
@@ -5,14 +5,6 @@ module GroupsHelper
     !person.member_of?(group)
   end
 
-  def group_open?(group)
-    !group.full?
-  end
-
-  def group_active?(group)
-    !group.inactive
-  end
-
   def show_member_buttons?(person, group)
     if !person_signed_in?
       false
--- a/app/helpers/groups_helper.rb
+++ b/app/helpers/groups_helper.rb
@@ -5,8 +5,12 @@ module GroupsHelper
     !person.member_of?(group)
   end
 
-  def even_a_button?(group)
+  def group_open?(group)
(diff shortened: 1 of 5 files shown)


app/helpers/groups_helper.rb
```diff
@@ -5,10 +5,14 @@ def show_join_group_button?(person, group)
     !person.member_of?(group)
   end
 
-  def even_a_button?(group)
+  def group_open?(group)
     !group.full?
   end
 
+  def group_active?(group)
+    !group.inactive
```
til: Couldn't this also use the variant with the question mark, `inactive?`? Active record usually adds both variants for boolean columns, and it's good practice to use the one with question mark where possible (a so-called predicate method).

til: Actually, now that I think of it, those are attributes that would fit well with the model, don't you think?

Previously, it was about deciding if there should even be a button, which is out of the model's responsibility, but saying whether it is active or not, or full or open, could definitely be done within the model itself.

til:
I'll miss you too, `even_a_button?` :sob:

app/helpers/groups_helper.rb
berlintam: yes! you're totally right. Sorry, I was stuck in helper-mode & trying to figure out what is even a button.

sareg0:
(!)[http://new3.fjcdn.com/gifs/Ron+swanson+actually+happy_6cf6f3_4786756.gif]

PR #441, opened 2016-03-16
Update deployment commands

lislis:
I got some errors when I tried to deploy, so this is mostly copy paste convenience


til:
I don't understand why bundle exec works better than bin/rake, but so be it.

lislis:
Technology, how does it work?

PR #442, opened 2016-03-16
update rails

lislis:
I bundle updated to see what happens.
@berlintam you wanna take it from here ;)


berlintam:
:scream:

lislis:
@berlintam awesome!
finger guns

berlintam:
thanks! I dunno if it'll break stuff thou... I mean... I clicked around & everything seemed ok but you never know. Should we just #yolomerge?

til:
I merged, should I deploy too or does somebody else want to?

Issue #326, opened 2015-04-28
Can't edit groups although I'm an admin


lislis:
reproduce:
- be an admin
- go to dashboard
- try to edit group

expected:
- groups form

actual:
- 403

berlintam:
Do we want admins to be able to edit groups? Seems a bit shady to me....

lislis:
I think for admins that makes sense. Don't you?

berlintam:
not really.... what would we even do in there?

roschaefer:
Admins don't have [group_member_powers](https://github.com/rubycorns/rorganize.it/blob/master/app/controllers/application_controller.rb#L29) for [action edit ](https://github.com/rubycorns/rorganize.it/blob/master/app/controllers/groups_controller.rb#L7) in the group's controller. Two solutions come to my mind:

1. Don't display the edit button in the admin dashboard. (e.g. ```if current_person.member_of?(group)```)
2. Lift the admins' rights to edit groups (e.g. ```ensure_member_powers``` means you are a member or an admin)

Opinions?

PragTob:
@berlintam I think it makes sense, e.g. if information is offensive or for whatever reason, admins should really be admin imo :)

PR #450, opened 2016-03-22
edit a group as an admin

berlintam:
closes https://github.com/rubycorns/rorganize.it/issues/326

And also @sareg0 and I used exciting html entities! http://htmlarrows.com/symbols/


berlintam:
next step, edit people?

app/views/groups/show.html.haml
```diff
@@ -3,6 +3,13 @@
     .col-md-12
       .page-header.clearfix
         %h1.pull-left= @group.name
+
+        .pull-right
+          - if current_person.admin?
+            %button.btn.btn-default.admin-button
+              = link_to edit_group_path(@group.id) do
```
til: The Rails path helpers should be able to deal with objects directly, and don't need an id. Everything is an object, everything should be able to deal with objects ...

app/views/groups/show.html.haml
```diff
@@ -5,7 +5,7 @@
         %h1.pull-left= @group.name
 
         .pull-right
-          - if current_person.admin?
+          - if current_person && current_person.admin?
```
til: Having to add checks like these are what makes programming unfun, but with the current way `current_person` works, it has to be. Wouldn't it be great if it wasn't necessary, and `current_person` would always return something? What could that something be? There is a pattern related to this, of which I can't remember the name nor find a good link right now, but let me tell you that it exists. Just food for thought, for a later improvement.

berlintam: hmmmm I don't quite get what you mean. 

Do you mean just create a helper method like `admin_signed_in?` & use that instead? Or is there an actual devise method that does this?

til: Yes, a helper method like that would be a great step in that direction!
With that, you would have nicely moved the tediousness of having to
check for current_person==nil into an isolated place, out of the haml
view. And as probably have heard from me already, the views should
contain as little logic as possible.

I don't know of an actual devise method that does exactly that, but it
should be possible to just add it as helper.

Another approach would be, to make sure current_person is always
something, never nil. This would be a variant of the NullObject pattern
(I think that's the name), unfortunately I still haven't found a good
link for it. But maybe devise gets in the way here, because it defines
current_person itself, and makes it harder for use to redefine it?

Anyway, just food for thought, none of this should stop you from merging
this PR if you want.

berlintam:
for others: this was helpful http://stackoverflow.com/questions/16077115/rails-replacing-try-with-the-null-object-pattern

Also, if you stub something by writing `allow(something).to_not receive(:stuff).and_return(things)` you get the following best error in the world:
```
 `allow(...).to_not receive` is not supported since it doesn't really make sense. What would it even mean?

**Comment by til** [comment]
Wohoo, great!

So there is a method `guaranteed_person` now, which can be called
instead of `current_person` - this introduces the feature gently,
without changing `current_person`. But to be honest, I think that
"guaranteed" sounds a bit weird from the point of view of the caller
(Why guaranteed? Should there be a corresponding method
`maybe_a_person`?). Did you consider changing `current_person` directly
instead - any reason why you think that could cause problems?

> Also, if you stub something by writing `allow(something).to_not
> receive(:stuff).and_return(things)` you get the following best error
> in the world:
>
>  `allow(...).to_not receive` is not supported since it doesn't really make sense. What would it even mean?

LOL!

**Comment by berlintam** [comment]
@til merge?

**Comment by til** [comment]
LGTM.

I can offer a few more small nitpicks if you want, but feel free to merge anyway.

**Comment by berlintam** [comment]
bring it

File: app/helpers/application_helper.rb [hunk 3]
```diff
@@ -25,4 +25,12 @@ def country_names_for_select(countries)
     countries.map { |c| [ISO3166::Country[c].name, c] }
   end
 
+  def current_person
+    @current_person ||=  if super.present?
+                              super
+                            else
+                              NullPerson.new
+                            end
```
til: Sooo, if super is nil it should not be used, right? Then maybe you could use the fact that nil can be used as a boolean too, and make this code shorter.

spec/models/null_person_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper'
+
+describe NullPerson do
+  let(:null_person) { NullPerson.new }
+
+  specify 'is not an admin' do
+    expect(null_person.admin?).to be false
```
til: The rspec world recommends to check for "truthy" and "falsey" instead. One reason is, that like this it _must_ be `false` and would fail if it was `nil` instead. However, in ruby they can be used interchangably in most cases.

spec/models/null_person_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper'
+
+describe NullPerson do
+  let(:null_person) { NullPerson.new }
```
til: You can also use `described_class` to avoid repeating the class name here. Not a big deal in this case, however in a large spec with many occurances of it, it makes it a bit more readable and easier to change the class name.

til:
Ran out of nitpicks.

app/helpers/application_helper.rb
berlintam: hmmm I don't quite understand. Are you suggesting then that I turn current_person into current_person? But this method should return a person object, not true/false, oder?

berlintam: Also if super is nil, then `super.present?` will return false, so it'll go to Null.Person.new, so... I guess the only way I can see to make this shorter is to use a ternary operator....

til: No, I was just trying to promote general ruby shortness: `super || NullPerson.new` should work here, instead of the if block.

berlintam: OooOOoOOoh! I get it! Thanks!

PragTob:
well done

lislis:
Sooooo, if I try to edit a group on production then I still get a 403. I deployed current master but it's not working. Is it the app or is it me?

til:
Same for me, I'm getting a 403 too. Are you sure the thin server was
restarted after the code was updated?

berlintam:
strange! is it working locally?

lislis:
I re-deployed and now it's working!
Not sure what I did wrong though, I followed all the steps in the Readme...
Anyway, yaaay!

berlintam:
omg!

PR #451, opened 2016-03-27
make group email accessible on admin dashbaord, closes #466

### LINKED ISSUE #466: Thumbnails on groups#index have the wrong size
We render the wrong thumbnail size for group images and then 'scale' them up via css which results in blurry images. No good. We have the appropriate size (as thumbnails on people shows you) so we should use it.

lislis:


berlintam:
:+1:  kewwwwwwl

til:
Cool, thanks!

Btw, there's a helper method for mailto links:

http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-mail_to

whoever feels like making the code _even better_, feel free to change
it!

PR #452, opened 2016-03-27
improve explanation of group email, closes #447

### LINKED ISSUE #447: Improve explanation of email field when creating a new group
Currently the form has this text next to the group.email field:

"Where we'll send official emails (only visible to us)"

and

"Please don't use a googlegroup because emails from us won't go through"

it would be helpful if it also said that Group Join notifications are sent to that email address. Maybe "automated" instead of "official"?

lislis:


berlintam:
:+1: woot!

til:
A late: woot! from me too

A remark regarding the "*" in the label: this should indicate that the
field is required, right? I think there is a way in Rails to get this
automatically, depending on the validations for the attribute in the
model, but I don't know exactly how. If anyone feels like diving deeper
into rails forms, this might be a nice opportunity.

PR #453, opened 2016-03-27
add male_student flag

lislis:
Hey there!
As we had the this discussion some time ago, I'll just did a simple implementation of an indicator if a group welcomes men as students.
Please let me now what you think.
I also added some explanations about our other flags.
screen shot 2016-03-28 at 00 56 52
screen shot 2016-03-28 at 00 56 42

### DIFF
--- a/app/views/groups/_form.html.haml
+++ b/app/views/groups/_form.html.haml
@@ -66,5 +66,9 @@
       .form-group
         = f.check_box :inactive
         = f.label(:inactive, "Is your group inactive?")
+      .form-group
+        = f.check_box :allow_male_students
+        = f.label(:allow_male_students, "Do you welcome male students?")
+        %p.help-block Many groups prefer to only have students who identify as women. Please indicate if you also welcome students who identify as men.
     .actions.pull-right
       = f.submit class: "btn btn-primary"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,9 +11,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160105185150) do
+ActiveRecord::Schema.define(version: 20160327223742) do
 
-  create_table "friendly_id_slugs", force: true do |t|
+  create_table "friendly_id_slugs", force: :cascade do |t|
     t.string   "slug",                      null: false
     t.integer  "sluggable_id",              null: false
     t.string   "sluggable_type", limit: 50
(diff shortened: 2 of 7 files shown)


app/views/groups/_form.html.haml
```diff
@@ -64,5 +64,9 @@
       .form-group
         = f.check_box :inactive
         = f.label(:inactive, "Is your group inactive?")
+      .form-group
+        = f.check_box :allow_male_students
+        = f.label(:allow_male_students, "Do you welcome male students?")
+        %p.help-block Most groups prefer to only have students who indentify as women. Please indicate if you also welcome men.
```
berlintam: perhaps instead "Please indicate if you also welcome students who identify as men"

db/schema.rb
```diff
@@ -11,9 +11,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160105185150) do
+ActiveRecord::Schema.define(version: 20160327223742) do
 
-  create_table "friendly_id_slugs", force: true do |t|
+  create_table "friendly_id_slugs", force: :cascade do |t|
```
berlintam: I guess this change was generated on its own. Any idea what it does & why?

berlintam:
woot woot! love it!

app/views/groups/_form.html.haml
berlintam: Also maybe 'many groups' instead of 'most groups'.

db/schema.rb
lislis: actually, no. I'll reset that

lislis:
I made the changes and rebased!

lislis:
I did have to run the migrations that changed the that one line again that @berlintam commented on earlier. But no idea what it does.

app/views/groups/_form.html.haml
```diff
@@ -64,5 +64,9 @@
       .form-group
         = f.check_box :inactive
         = f.label(:inactive, "Is your group inactive?")
+      .form-group
+        = f.check_box :allow_male_students
+        = f.label(:allow_male_students, "Do you welcome male students?")
+        %p.help-block Many groups prefer to only have students who indentify as women. Please indicate if you also welcome students who identify as men.
```
til: "indentify" *chuckle*

til:
LGTM expect for the unrelated change in schema.rb.

lislis:
@til could it be that the change in the schema is related to some newer version of something? When I run the migrations it just happens?

til:
Yes, apparantly it is caused by an update of rails:

http://stackoverflow.com/questions/27697119/what-is-cascade-in-rails-schema-rb-and-where-did-it-come-from

I think your version, the one with :cascade, is the newer one, and since
this is the first time that schema.rb is changed since the rails update,
this gets updated as well.

So it is ok to update it I'd say in this case. Usually I would advise to
put changes like that in a separate commit, because it is not related to
the functionality that you are adding here, but I don't see how this
would be possible in this case, because it is only triggered when the
schema is changed.

lislis:
I separated the migration and schema into its own commit. I can't really pull them apart any further. I also rebased. Please merge!

berlintam:
https://media.giphy.com/media/Q1D81ZTkMxQ1W/giphy.gif

lislis:
nod

PR #454, opened 2016-04-04
use mailto helper

lislis:
... instead of hacking the link_to helper


berlintam:
👍

sareg0:
there are a few more 'mailto' links floating throughout the code, I will update them too with the helper, and then we can merge this puppy

Issue #449, opened 2016-03-22
Joining a group :sparkles: a draft :sparkles:


lislis:
OK, so here is the proposed workflow:
- user goes on the group page, there is only button that says "join the group"
- the user clicks the button and a message field opens with a default message like `I want to join your group as student`. the user can edit the message. There is also an explanatory text that says that the message will be send to the group for approval. And there is a "request membership" button.
- the user clicks the "request membership" button. The message field closes with a notice "Your request was sent to the group." and there is now a "membership pending" indicator instead of "join the group". 
- In the background an email is sent to the group_email and the pending membership is shown. On the group page as pending membership. duh.
- a member of the group can (un-) approve the membership. (nice to have: send an email to the new member)

til:
I like it!

However I want to warn, that this is not a trivial feature, so don't
expect to be able to implement it in half an hour ;-)

Also, it would be a great opportunity to practice developing guided by a
feature spec (aka integration test, aka end-to-end test, aka functional
test). Part of it is sending out the emails. Check out topic 10.3 in
http://guides.rubyonrails.org/testing.html .

With ActionMailer::Base.deliveries.last it should be possible to access
the latest sent email, and check its contents, somehow like this:

join_notification = ActionMailer::Base.deliveries.last
expect(join_notification.subject).to eq('Erna Example wants to join the Rubycorns')

berlintam:
cool!

Can any of this be broken down into smaller tasks? Then we can all work on it together! #teamwork #rubycornsrule

sareg0:
We closed this PR just to keep things nice and clean: https://github.com/rubycorns/rorganize.it/pull/456

However, we can re-open it if needed.

lislis:
I'm tackling this now in PR #648

PR #456, opened 2016-04-19
WIP - joining a group feature via BDD

BriocheBerlin:
As the title would suggest, we (1: 'til, 2: 'sara', 3: 'brigitte') have started this feature and are going ground up with BDD. We will continue this for the next few weeks. Others who would like to follow this 'theme' with us should come and work on it with us every Tuesday. We will be 'pairing' and working on it all together!

### DIFF
--- a/spec/features/group_join_spec.rb
+++ b/spec/features/group_join_spec.rb
@@ -1,0 +1,29 @@
+require 'spec_helper'
+
+feature 'User joins a group', :vcr => {:cassette_name => "create_group" } do
+  scenario 'join a group' do
+    visit_group_page
+    click_join_button
+    edit_default_message
+  end
+
+  # let(:person) { create(:person) }
+  # let(:group) { create(:group) }
+
+  def visit_group_page
+    create(:group, name: 'rubycorns')
+    person = create(:person)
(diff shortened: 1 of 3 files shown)


berlintam:
exciting!  Are you just going to re-write the current join a group behaviour or does this also include all the pending membership stuff?

spec/features/group_join_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+feature 'User joins a group', :vcr => {:cassette_name => "create_group" } do
```
til: Extra bonus points to whoever removes the necessity of vcr here! It's currently needed because creating a group triggers geocoding automatically.

til:
The full plan for the feature also includes membership pending and
everything, as @lislis outlined it all in #449. Maybe on the way an
intermediate version emerges that can be deployed already before
finishing everything, we'll see.

So yes, this is about all of #449. Worth mentioning it again, because we
forgot to link it before from this PR.

PragTob:
anything moving forward here? :)

til:
Would anybody be offended if we close this?

3 ... 2 ...

PR #459, opened 2016-04-26
fixes #455 incorrect error msg bug

### LINKED ISSUE #455: deleting a person from a group results in incorrect error message
A few weeks ago I created a second Tam account on Rorganize & joined the cornz just to double check that 'a person has joined your group' emails were getting sent out. Today I deleted that Tam from rorganize, and upon doing so, I was redirected to /groups with the error message 'you have left your group, please consider buying everyone cake on your last day' or whatever is it. (I double checked and I definitely removed the correct Tam and I am still a member of rubycorns). 

So.... that's a bug.

berlintam:


app/controllers/memberships_controller.rb
```diff
@@ -15,16 +15,21 @@ def create
...
+
     flash[:success] = leave_message
     redirect_to groups_path
   end
 
   private
 
   def leave_message
-    "You have left your group. No doubt they will miss you." +
+    if @membership.person_id == current_person.id
```
til: It should be possible to compare the actual persons with each other whether they are equal, instead of their ids. In general, I'd say dealing with the actual object is preferred in ruby, over using ids or other "technical details".

spec/controllers/memberships_controller_spec.rb
```diff
@@ -42,25 +41,47 @@
...
   describe 'destroy' do
 
-    before do
-      person.join!(group)
-      @membership = person.memberships.first
-    end
+    context 'removing yourself from a group' do
+      before do
+        person.join!(group)
+        @membership = person.memberships.first
```
til: I know, it was like this before, but if you are interested in improving this spec, you might want to consider using a `let` thingy here instead of an instance variable. I don't think there is any specific reason why an instance variable would be needed, and it's more common to use `let` within rspec.

berlintam: 👍

app/controllers/memberships_controller.rb
berlintam: ah, cool! will do!

PragTob:
Something moving forward here? :)

berlintam:
oh gawd I forgot about it.

PragTob:
sobbing

berlintam:
sorry

PragTob:
did i miss it

berlintam:
here you go @PragTob. sorry this took so long!
sorry

PragTob:
No worries, thanks a lot!

thanks

PR #462, opened 2016-06-07
Divide groups into active/inactive #index, close #436

### LINKED ISSUE #436: Don't show inactive groups
I made a list with RGB Groups that don't exist anymore/at the moment. Would be great, if they can be taken out of the list:

Rubies in the Woods
Wednesday Ruby
Coding Cats
Rubies on da Cloud
RubyDoobyDoo

lislis:
List active groups first, then inactive groups.

Does that make sense?

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -7,9 +7,13 @@ class GroupsController < ApplicationController
   before_action :ensure_member_powers, except: [:index, :show, :new, :create, :destroy]
 
   def index
-    @groups = Group.order :name
+    @groups_inactive, @groups = Group.all.order(:name).partition { |g| g.inactive? }
+    
     @groups = @groups.by_country(params[:country]) if params[:country].present?
     @groups = @groups.by_city(params[:city]) if params[:city].present?
+
+    @groups_inactive = @groups_inactive.by_country(params[:country]) if params[:country].present?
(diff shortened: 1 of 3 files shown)


til:
It makes sense, and looks to me like it solves the issue well. However, there is a lot of duplication of code in the view. I didn't compare every detail, but isn't most of the view for active and inactive groups the same? One way to DRY this up would be to put the similar code in a partial, and render that, e.g.:

```haml
- @groups.each do |group|
  = render partial: 'one', locals: { group: group }
```

Hmm, I think there is an even shorter way, that I haven't used for a long time. Untested:

```haml
= render partial: 'one', collection: @groups
```

lislis:
Very nice! I updated the code

app/controllers/groups_controller.rb
```diff
@@ -7,11 +7,11 @@ class GroupsController < ApplicationController
   before_action :ensure_member_powers, except: [:index, :show, :new, :create, :destroy]
 
   def index
-    @groups = Group.where(inactive: false).order :name
+    @groups_inactive, @groups = Group.all.order(:name).partition { |g| g.inactive? }
+
```
til: Good idea!

til:
Looks good! :+1:

lislis:
Rebased and squashed

til:
Thank you!

lislis:
thanks!

til:
deployed

PR #463, opened 2016-06-21
fix double scrollbar, fixes #362

### LINKED ISSUE #362: Double vertical scrollbar
In Firefox, I see two scrollbars on the right border, of which the rightmost one is without functionality, and instead the left one must be used to scroll the page content vertically. The left one looks wrong though. See attached screenshot.

Since we are using a fixed navbar from bootstrap, I think it should instead look like in this example: http://getbootstrap.com/examples/navbar-fixed-top/

It has something to do with the overflow styles that are set on the body. Why are they set on the body? Can we try to remove them?

ror-scroll

lislis:


til:
Hooooorrrraaaaaayyyy!!!!

Phew, seriously, thanks. It has annoyed me forever when looking at the site, making me feel like I should be sorry that I use Firefox ...

lislis:
\o/

PR #464, opened 2016-06-21
[WIP] Comments for topics, close #104

### LINKED ISSUE #104: Users can comment on topics
As a user on the topics show page
I want to be able to comment on a topic
so I can ask some questions or add some information.

lislis:


lislis:
So, a few question came up:
In the comment model, can I say `belongs_to :group, through: topic` or something?

For now I'd like to just display comments underneath the respective topics. We get topics in `group#show`. Would I get comments for each topic here? Feels kinda strange

db/migrate/20160621202750_create_comments.rb
```diff
@@ -0,0 +1,11 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.string :user_name
+      t.text :body
+      t.integer :topic_id
+
+      t.timestamps
+    end
```
til: In a Rails migration, a created column allows NULL by default, so all of these allow NULL. I suggest to consider for each column, whether a NULL value makes sense, and what it would mean. E.g., NULL in `topic_id` would mean - what? A lost comment? Where it doesn't make sense, I suggest to add `null: false`.

til: By the way, those constraints are different from the validations. It's a common misunderstanding in Rails what the purpose of validations vs. database constraints are, and traditional Rails tended to neglect the latter. However, I think database constraints are definitely useful.

til: Did you mean to add a `person_id` instead of `user_name`? The `belongs_to :person` in Comment definitely would need it. The question is, whether only logged in users should be able to comment (I suppose yes), then yes, this would need a `person_id`.

til:
When the comments are displayed next to each topic, then yes, it would make most sense to get them through the topic. Doesn't feel strange to me - something like this (pseudo haml):

```haml
- @topics.each do |topic|
  %h3= topic.text
  - topic.comments.each do |comment|
    %p= comment.body
```

When not doing anything special, this will result in one database query for each topic, which can become inefficient when there are many topics on one page. But there is a way around it, with - I think - an `:include` option when loading the topics. I'd start without this first, and then try to add the option later.

lislis:
So, after some back and forth and troubles with displaying the comment form, I thought of a different UI solution. 
We currently don't have a `topics#show` view because we only show topics in `groups#show`. I'd like to add `topics#show` and use that to display the comments and a comment form.
At the same time I'd clean up topic section in `groups#show`. wdyt?

app/views/topics/index.html.haml
```diff
@@ -21,6 +22,10 @@
             %ul.dropdown-menu{role: 'menu'}
               %li= link_to 'Edit', edit_group_topic_path(topic.group_id, topic)
               %li= link_to 'Destroy', group_topic_path(topic.group_id, topic), method: :delete, data: { confirm: 'Are you sure?' }
+      .row
+        - if topic.comments
```
til: Usually an association like this can not be nil - if it is, then something else went wrong and should be fixed there. So I think the defensive if check is not necessary here.

app/controllers/groups_controller.rb
```diff
@@ -49,6 +49,9 @@ def show
 
     @topic = Topic.new
     @topic.group = @group
+
+    @comment = Comment.new
+    @comment.person_id = current_person if current_person
```
til: Isn't current_person always defined, because we (or rather Tam) added an Anonymous user?

Besides that, in Rails associations a new model that is already associated can also be built like this: `current_person.comments.new`.

app/controllers/comments_controller.rb
```diff
@@ -6,6 +6,13 @@ def edit
   end
 
   def create
+    @comment = Comment.new(comment_params)
+
+    if @comment.save
+      redirect_to @comment.topic.group, notice: 'Comment posted!'
+    else
+      redirect_to @comment.topic.group, notice: 'Whoopsie!'
```
til: Or maybe not redirect but render the view again, so that the validation errors are displayed? Otherwise after the redirect, the current`@comment` that contains the error messages would be lost again.

til:
Do you mean that the added `topics#show` will be displayed on a
separate page then? What about `groups#show`, would you remove the
topics from there, or show them as well?

lislis:
I updated things. There is now `topics#index` which is nested in groups. And topics#show should display its comments. But I still need to make the form partials work for both topics and comments...

lislis:
The logic it there now. Next up tests! I'll leave the templates in their basic state looking at the redesign happening

lislis:
So, I added some basic tests. Let me know if there should be more. Or rather, what I should test for?
It's also rebased on current master, but not yet squashed.
ping @til for a review :D

lislis:
So, I added some basic tests. Let me know if there should be more. Or rather, what I should test for?
It's also rebased on current master, but not yet squashed.
ping @til for a review :D

lislis:
So, I added some basic tests. Let me know if there should be more. Or rather, what I should test for?
It's also rebased on current master, but not yet squashed.
ping @til for a review :D

app/assets/stylesheets/custom/custom.css.scss
```diff
@@ -6,7 +6,7 @@
 // other styles
 @import "_layout.scss";
 @import "_dashboard.scss";
-@import "_posts.scss";
+@import "posts";
```
til: Unrelated change? Or is this necessary?

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    else
+      redirect_to group_topic_path(@comment.topic.group, @comment.topic), notice: 'Something is wrong, no cake yet!'
+    end
+  end
+
+  def update
+  end
+
+  def destroy
+  end
```
til: Comments can only be created, right? Then all the other actions in this controller can not be reached, and the methods should be removed to avoid confusion.

app/controllers/topics_controller.rb
```diff
@@ -1,17 +1,30 @@
...
   before_action :authenticate_person!
 
   def edit
   end
 
+  def index
+    topics = Topic.where(group_id: params[:group_id])
+    @covered_topics, @future_topics = topics.partition { |topic| topic.covered? }
+    @group = Group.where(id: params[:group_id]).first
+    @topic = Topic.new(group_id: params[:group_id], person_id: current_person.id)
```
til: In most cases in Rails, when you have an instance of a model, it can be used directly instead of the id, and should, in my opinion, e.g. in this case it could be `person: current_person`. This fits better to what is happening - creating a Topic that is related to a Person. That the relation is constructed via an id is a technical detail that doesn't need to be shown explicitly.

app/controllers/topics_controller.rb
```diff
@@ -1,17 +1,30 @@
...
+    @comment = Comment.new
+  end
+
   def create
     @topic = Topic.new(topic_params)
     @topic.person = current_person
 
     if @topic.save
-      redirect_to @topic.group, notice: 'Topic was successfully created. If you bring cake, people
+      redirect_to group_topics_path(@topic.group.id), notice: 'Topic was successfully created. If you bring cake, people
```
til: Similar thing here: `group_topics_path(@topic.group)` should work too.

app/views/topics/show.html.haml
```diff
@@ -1,2 +1,36 @@
...
+      =link_to  @topic.person, person_path(@topic.person)
+      on
+      = @topic.created_at.strftime("%d %B %Y")
+      -if @topic.covered_at
+        %p
+          covered on
+          = @topic.covered_at.strftime("%d %B %Y")
+
+  .topic-comments
+    - if @comments.size > 0
```
til: In Ruby there is e.g. this which makes it even more readable: `if @comments.any?`

db/migrate/20160621202750_create_comments.rb
```diff
@@ -0,0 +1,11 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.string :person_id
+      t.text :body
```
til: Does it make sense when either of these fields is NULL? I doubt it, and recommend to declare them as NOT NULL (`null: false`) as well. This way it is sure there will never end up a misleading NULL value in the database.

til:
This looks great already!

Some minor comments from me inline.

Regarding specs - since the main thing that makes up this feature is to
add new comments, and this is fully covered in a feature spec, I think
that would be sufficient. I am not sure about the usefulness of the
routing specs - did we have similar ones already? Oh well. They could
probably all be removed imho.

One minor thing on the feature spec - it is checking for a text that is
set in the factory ("Let's talk about making and eating cake"). It is
better if all texts that are checked for are also visible in the feature
spec file, just so that all is in one place and can be seen more or less
at once. So I suggest to move that text from the factory to the feature
spec.

Theoretically you could add a spec that checks for the case of a
validation error, when someone tries to post a comment that is empty,
and make sure that no comment is posted in this case. But I don't think
it is essential.

app/controllers/topics_controller.rb
lislis: So, I could remove some of the `.id` occurences, but especially when they are used in path helpers I get error messages. I think this has to do with the slugify group names feature?

app/controllers/topics_controller.rb
lislis: I'm running into errors here. interesting is that this gets me to `/group/id/topics` whereas leaving out the id gets me to `/group/slug/topics` which then throws an error. Is there something with group slugs that we need to adjust?

db/schema.rb
```diff
@@ -11,13 +11,21 @@
...
-ActiveRecord::Schema.define(version: 20160327223742) do
+ActiveRecord::Schema.define(version: 20160621202750) do
+
+  create_table "comments", force: :cascade do |t|
+    t.string   "person_id"
+    t.text     "body"
+    t.integer  "topic_id",   null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
```
til: The `null: false` changes in the migration are not reflected here yet. To fix that, check out master, do a `rake db:reset`, then check out this branch, then `rake db:migrate`. This should re-apply the migration, and change the schema.rb file, which can then be committed.

til:
Very nice. The new feature spec that tests the validation failure looks
good.

Could very well be that the path helper errors are caused by something
of the slugification. I'd say don't let this stop you from merging this
PR, it can and should be fixed separately.

Apart from the not yet updated schema.rb, :+1:

lislis:
Lol, I totally forgot to run the migration again :D

lislis:
@til should I squash and rebase? Or is anything else missing?

til:
Please go ahead. With the :+1: I meant feel free to merge!

lislis:
Wheee
yay

til:
Cool! Deploy?

lislis:
deployed with blog post!

til:
Hm, when I click on "Topics this group talks about" I get the message
"You are not allowed to do that", anyone else seeing this too? Both when
I'm logged in and when not.

til:
screenshot_20160915_195941

lislis:
@til are you a rubymonstas member?

lislis:
Well, I can see from the screenshot that you are not :D 
You have to be a member to see topics. But now that you ran into this issue, maybe that was a sub-optimal decision?

til:
Maybe, I think it's confusing. I can't remember or find any discussion
about this - what is the reason for not allowing non-members to see the
topics?

If hiding the topics is on purpose, then I think at least the link
should be hidden, so users don't get an unexpected "You are not
allowed" message.

lislis:
Discussion continues here https://github.com/rubycorns/rorganize.it/issues/481

PR #469, opened 2016-07-13
Fix bug filtering groups

### LINKED ISSUE #465: Filter groups is broken on production
Reproduce:
Go to groups and try to filter by city and or country.

Result:
500 error

Should:
Show filtered list

Note: @berlintam noticed that filtering on people works. Strange thing is, filter on people has a reset button, groups doesn't but should. It is a mystery ???

veracl:
This fixes #465 (makes the reset button show up as well)

The error message when I tried recreating the bug locally was "undefined method `by_city' for #<Array:...>", so I assume it was caused by the following:
"partition" returns an array, and by_country/by_city are scopes which can only be called on ActiveRecord::Relation objects.
Please correct me if I am wrong :)

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -7,8 +7,9 @@ class GroupsController < ApplicationController
   before_action :ensure_member_powers, except: [:index, :show, :new, :create, :destroy]
 
   def index
-    @groups_inactive, @groups = Group.all.order(:name).partition { |g| g.inactive? }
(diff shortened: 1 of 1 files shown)


til:
This sounds plausible. Great explanation, and thanks for the fix!

til:
Deployed, now the 500 is gone.

lislis:
amazing, thanks @veracl!
finger guns

PR #471, opened 2016-07-26
Params for redesign.

### LINKED ISSUE #470: Staging Server for Redesign
In doing the redesign, it would be great for @katrin-k to have a place to deploy her changes to, where we could take a quick look, without having to open her branch locally every time.

I propose a staging server, but want to know if this is even feasible with the hosting that we have.

If not this as a solution. @til you mentioned another way; a route like '/redesign' where the new design would live. I felt like it wouldn't be that simple to do, but please let me know if I am mistaken.

If you could all help me with some brainstorming and a solution for Katrin, that would be great!!

BriocheBerlin:
Adds params redesign for @katrin-k's design overhaul.

To see future changes in the desing add changes to `redesign.css.scss`, then add "?redesign=true" to url, e.g. "https://rorganize.it/coaches?redesign=true".
(Currently it just completely breaks the desing and colors the background purple. Because I like purlple.)

Closes #470.

### DIFF
--- a/app/assets/stylesheets/redesign.css.scss
+++ b/app/assets/stylesheets/redesign.css.scss
@@ -1,0 +1,3 @@
+body {
+  background-color: purple;
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -3,7 +3,7 @@
 <head>
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <title>RoRganize.it | Find a Rails Girls Study Group</title>
-  <%= stylesheet_link_tag    "application", media: "all", "data-turbolinks-track" => true %>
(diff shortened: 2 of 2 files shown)


til:
Cool! Next step: deploy, then move @katrin-k's design changes from 468-redesign into this scheme, so that it can be previewed live.

berlintam:
super cool!

til:
By the way, as an effect of #472 this is deployed now. Look here: https://rorganize.it/?redesign=true

berlintam:
it's ummm.... interesting!

sareg0:
how awesome! Thanks for doing this @BriocheBerlin

BriocheBerlin:
didn't change much really. (although it took me some time... but @til was patient as always 😸 )

PR #476, opened 2016-08-09
468 Redesign pt. one

katrin-k:

PR #478, opened 2016-08-10
Initial redesign CSS for merge into master

katrin-k:
Here's the redesign CSS file, as mentioned in closed PR #476 // cc @til .


til:
Deployed. Works!

https://rorganize.it/?redesign=true

One caveat is that when clicking on any link, the parameter is gone and needs to be appended manually again. I guess that's ok - dunno if there is a good solution around it.

PR #479, opened 2016-08-11
Set redesign param for all links

katrin-k:
If the URL contains the string "redesign", all links on the page will get the param `?redesign=true` added to their hrefs.


til:
Deployed, works, cool!

Issue #468, opened 2016-07-12
Redesign + refactoring


katrin-k:
@lislis, @sareg0 and me talked about how to implement the new design. We'll have a new directory under http://rorganize.it/redesign. 

There we can work on 
- refactoring scss files
- implement the new layout
- work on UX issues (until a certain point that needs tbd)

At some point, launch all this

sareg0:
@katrin-k when you make your changes to your files and they are deployed, they will be viewable via https://rorganize.it/?redesign=true thanks to @BriocheBerlin and her brilliant PR: 
https://github.com/rubycorns/rorganize.it/pull/471

Here is how your workflow should look.

Use [application.css.scss](https://github.com/rubycorns/rorganize.it/blob/master/app/assets/stylesheets/application.css.scss) as a guide. In this file we `require` the files/folder we are going to use in the app.
Include all of your files/folder in the new [redesign.css.scss file](https://github.com/rubycorns/rorganize.it/blob/master/app/assets/stylesheets/redesign.css.scss)

Put these in a folder called `redesign` or something of the like (if you haven't already), and place this folder inside the `[app/assets/stylesheets` folder](https://github.com/rubycorns/rorganize.it/tree/master/app/assets/stylesheets).

You can push your changes and make pull requests, and as long as you have required the appropriate files and folders in `redesign.css.scss`, you will see them live via https://rorganize.it/?redesign=true

Do ping us as you go, and we can help you if you want to pair/need advice/etc.

til:
Lots of changes already in the redesign branch, very cool! I'm very curious to see how it is going to look!

It would be great if we can make all those changes previewable to us and others via https://rorganize.it/?redesign=true (right not this looks still a bit, uhm, basic). For that to work, two parts are needed: first, the switching of the css files, which is already accomplished in the master branch (also thanks to @sareg0's detailed instructions!) Second, a way to have two different version of templates (`.html.haml`) files in parallel, since those change as well in the redesign.

We could use a feature of rails called Action Pack Variants for this. http://edgeguides.rubyonrails.org/4_1_release_notes.html#action-pack-variants . Basically it would allow to add a redesigned variant of a template next to the original version, like this:

```
app/views/coaches/index.html.haml
app/views/coaches/index.html+redesign.haml
```

and Rails would automatically use the redesign variant when called with `redesign=true`. What do you think about this - anyone wants to give it a try? (I'll be at Asquera tomorrow and happy to help in person).

[...]

PR #484, opened 2016-09-26
Redesign: finish rewriting all pages

katrin-k:
If this is merged, the testing and bugfixing and last 20% percent can begin (see also #468)


til:
Wow, so many templates!!!

I glanced over it briefly - none of this should currently affect the
live site with the old design, so it can safely be merged and deployed,
right? Except for makeGif being commented out (is that needed?), and the
javascript in redesign-main.js, which may or may not affect the old
pages.

app/assets/javascripts/redesign-main.js
```diff
@@ -0,0 +1,15 @@
...
+// 2. When tab link is clicked
+$('.group-tabs a').click(function(e){
+  e.preventDefault;
+  // 2.1 Change tab link status to "active"
+  $('.group-tabs a').removeClass('active');
+  $(this).addClass('active');
+  // 2.2 Show tab corresponding to the link
+  $('.tab').hide();
+  $($(this).attr('href')).show();
+});
```
til: All of this should be executed only after the document is ready. The way it is written now, it will be executed immediately when the JavaScript code is loaded, which can be a problem: when the elements that it refers to (e.g. `$(".tab")`)are not loaded yet, then the selector will not match anything and the code does not do what is expected.

There is a jQuery shortcut to run code after document ready, see main.js for an example.

Also, maybe there is a way to only run this when in the redesign mode?

katrin-k: Aaaah... of course! Thanks for the hint.

katrin-k:
@til Correct, all of the HTML pages have the `…+redesign…` addition.    
Regarding the redesign-main.js: I will double-check to see it doesn't interfere with any of the Bootstrap classes. 
Regarding makeGif: Naaa, I can take it in again. Though indeed, it's not necessary for functionality. (I was just commenting it out since it was using my CPU too much.)

til:
Sorry I didn't say that earlier: besides the two minor comments I made, this can be merged and deployed. The earlier the better, so we can all look at it on the live site (yay!).

til:
And usually our policy for merging is that the author of the PR merges themselves, after getting approval from someone else of the group.

app/assets/javascripts/redesign-main.js
katrin-k: Indeed, seems to work with .ready(). However, I've outcommented it for now. The script for appending the redesign param to all hrefs gets in the way of setting the anchor link. But this can also be done later.

katrin-k:
Alright, indeed I wasn't aware of the merging protocol ;).

til:
Cool, thanks! I deployed it now.

https://rorganize.it/?redesign=true

PR #485, opened 2016-09-29
Adds deamon to autostart thin server if down

### LINKED ISSUE #480: add autostart script to uberspace
Hey! Sorry for the awkward wording....
The following https://twitter.com/ubernauten/status/774061796533559296

Uberspace rebooted our server because of a RAM upgrade. That terminated our thin server.
All is fine, I restarted it manually, but it would be cool if we had a script that executes thin if the host machine reboots.

BriocheBerlin:
Has to get symlinked into service folder on uberspace.
Closes #480

### DIFF
--- a/deploy/autostart
+++ b/deploy/autostart
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+# start thin as a process in the foreground
+# (when thin is down, deamontools will execute)
+bundle exec thin -a 127.0.0.1 -p 34567 start
+echo "autostarting thin app server"


deploy/autostart
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+# start thin as a process in the foreground
+# (when thin is down, deamontools will execute)
+bundle exec thin -a 127.0.0.1 -p 34567 start
+echo "autostarting thin app server"
```
til: The line with the `echo` call will not be executed as long as thin is running in the foreground. Maybe swap the order of the lines?

til:
Cool, thanks for taking care of this!

Do you want to install the symlink and see if it works? I think one needs to create a directory and then symlink the script as `run` from there. There is already one service in our accout for postgresql, which can serve as example, here is how it looks currently:

```
[ror@schedar ~]$ ls -lah service/
total 8.0K
drwxr-xr-x.  2 ror ror 4.0K Jan 26  2016 .
drwx------. 18 ror ror 4.0K Sep 29 23:07 ..
lrwxrwxrwx.  1 ror ror   28 Jan 26  2016 postgresql -> /home/ror/etc/run-postgresql
[ror@schedar ~]$ ls -lah service/postgresql/
total 20K
drwxr-xr-x. 4 ror ror 4.0K Feb  2  2016 .
drwxr-xr-x. 3 ror ror 4.0K Feb  2  2016 ..
drwxr-xr-x. 4 ror ror 4.0K Jan 26  2016 log
-rwxr-xr-x. 1 ror ror  293 Feb  2  2016 run
lrwxrwxrwx. 1 ror ror   28 Feb  2  2016 run-postgresql -> /home/ror/etc/run-postgresql
drwx------. 2 ror ror 4.0K Sep  9 03:39 supervise
[ror@schedar ~]$ cat service/postgresql/run
#!/bin/sh -e
export USER=ror
export HOME=/home/ror
. $HOME/etc/postgresversion
export PATH=/package/host/localhost/postgresql-${POSTGRESVERSION}/bin:$PATH
export LD_LIBRARY_PATH=/package/host/localhost/postgresql-${POSTGRESVERSION}/lib/:$LD_LIBRARY_PATH
exec postgres -D $HOME/postgresql 2>&1
```

Hm, but `run` in this case is not a symlink but an actual file - it could be that daemontools wants it like that. So then your script should be copied instead of symlinked. Also, there is a symlink to a directory, `run-postgresql`, this looks like a mistake to me, but I am not sure. Do you want to have a look at it?

PR #490, opened 2016-11-15
add uptime script

lislis:
for much fun!

### DIFF
--- a/scripts/uptime.sh
+++ b/scripts/uptime.sh
@@ -1,2 +1,1 @@
-curl -f https://rorganize.it > dev/null 2>&1 || echo "ALERT ALERT" |
-        mailx -s 'Rorganize is down' roar@rorganize.it
+curl -f https://rorganize.it > /dev/null 2>&1 || echo "ALERT ALERT" | mailx -s 'Rorganize is down' roar@rorganize.it
--- a/scripts/uptime.sh
+++ b/scripts/uptime.sh
@@ -1,0 +1,2 @@
+curl -f https://rorganize.it > dev/null 2>&1 || echo "ALERT ALERT" |
+        mailx -s 'Rorganize is down' roar@rorganize.it


berlintam:
wooooow! impressive!
@lislis can you explain like.... what this means? Like... what the words... mean? & how it works?

scripts/uptime.sh
```diff
@@ -0,0 +1 @@
+curl -f https://rorganize.it > /dev/null 2>&1 || echo "ALERT ALERT" | mailx -s 'Rorganize is down' roar@rorganize.it
```
lislis: `curl` requests a page via HTTP and returns whatever it returns (try it in your console). Because we don't care about the output in stdout, we move it over to `/dev/null` which is computational limbo. `2&>1` moves the stderr to the same location. 
the `-f` flag is for failing: if the status code of  the request is anything but 200 (like 404 or a 50something) we go to the other branch after the `||` which is sending an email with 'ALERT ALERT' as body and 'Rorganize is down' in the subject line to the mailinglist

lislis:
@berlintam I added some explanation

til:
Wohoo! An oldschool cron+bash script that does the job, well done!

Is it really working? There seems to be no explicit cron log on
Uberspace, so we can't check that. But there is the access log, and
checking to see when we got a request from curl, it says we get one
every 5 minutes which is exactly as expected:

[ror@schedar ~]$ grep curl logs/access_log|tail -n 5
185.26.156.0 - - [15/Nov/2016:22:05:03 +0100] "GET / HTTP/1.1" 200 3651 "-" "curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.21 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2"
185.26.156.0 - - [15/Nov/2016:22:10:01 +0100] "GET / HTTP/1.1" 200 3651 "-" "curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.21 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2"
185.26.156.0 - - [15/Nov/2016:22:15:03 +0100] "GET / HTTP/1.1" 200 3651 "-" "curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.21 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2"
185.26.156.0 - - [15/Nov/2016:22:20:03 +0100] "GET / HTTP/1.1" 200 3651 "-" "curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.21 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2"
185.26.156.0 - - [15/Nov/2016:22:25:03 +0100] "GET / HTTP/1.1" 200 3651 "-" "curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.21 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2"

To really really test it we would have to take down the site, but who
wants to do that.

til:
And as a last remark, to document our thoughts behind this: in case you
are wondering, yes, this does not protect against the whole Uberspace
server going down, because then the cron job would stop running as well.
Theoretically we could protect against that by making the uptime.sh
script call deadmanssnitch - whoever wants to implement that feel free
to.

We also tried to find a simple free pingdom alternative, and I'm sure
there are a bunch out there on the internet, but it was more fun and
better learning experience to write a handmade script.

PragTob:
I agree that this was a better learning experience, and that sometimes it's good to just KISS (Keep it simple, stupid) and thanks for the work!

It's also way better than what we had before, still tend to not trust monitoring scripts running on the same server. E.g. not only because the whole server might go down, but other smaller parts might break. E.g. Email sending might break, so that we don't get the notification email, or it might still answer to HTTP requests but due to some misconfiguration/network errors not to requests from the wider Internet.

--> still feel a little unsafe

Issue #483, opened 2016-09-26
Add "change password" option


katrin-k:
On the user profile page, I've added a button to "Change the password" (see https://github.com/rubycorns/rorganize.it/commit/9162c39936064aa13c53bfddfda3f624e0aca270). However, I was not able to figure out the correct url_path. 

It might also make sense to only show this option if the account is not linked with Github?

til:
Rails has a rake task that shows all routes of an application:

bin/rake routes

Each line is a route, and the column of the left has the corresponding
path helpers part, but without the _path or _url ending.

Searching for "password" in this output

bin/rake routes|grep password

returns among other things "edit_person_password", so maybe
"edit_person_password_path" is the right one? I didn't try it though.

And yes, only showing that when not connected through GitHub sounds
sensible to me.

raluka:
In current implementation, we allow an user to change password only through devise (as request provided via email). We can add option to change password in `edit` form for user, or in separate form. Also, we need to update controller to know about that (either `update_user` method, or create a separate method `change_password`)

katrin-k:
@raluka \o/ thanks!  Changing pw works sweetly. 
However, I do have a little problem finding the corresponding html (it needs adjustment to the new design). The routes table hints at `edit_person_registration_path 	GET 	/people/edit(.:format) 	registrations#edit`. But this… is not there? Yet… it does serve an html?

til:
This is an interesting question. We were looking for the corresponding
view as well yesterday, and didn't find it. Maybe they are still inside
the devise gem and need to be re-generated, following the procedure at:

https://github.com/plataformatec/devise#configuring-views

- maybe you want to try that?

We would have liked to remove the field for email, because we don't want
to allow changing the email for now, so if you find the view, please
remove it or let us know.

katrin-k:
Booyaka! 
This was indeed the reason. And: nice gem, worked pretty easy. Thanks for the link, @til!
I've only run it for the `registrations` and converted the generated file to haml. Will adjust the html and send the PR then.

katrin-k:
Redesign is done. This can be closed.

PR #491, opened 2016-11-22
Add action for Change Password link 

raluka:
[Add "change password" option #483](https://github.com/rubycorns/rorganize.it/issues/483)

This PR implements correct path for `Change password` button and makes change password works.

### DIFF
--- a/spec/features/update_password_spec.rb
+++ b/spec/features/update_password_spec.rb
@@ -1,0 +1,19 @@
+require 'spec_helper'
+
+feature 'update password' do
+
+  before { visit new_person_session_path }
+  before { sign_in person }
+
+  let(:person) { create(:person, password: 'password') }
+
+  scenario 'update password' do
+    visit person_path(person, redesign: true)
+    click_on 'Change password'
+    fill_in 'Current password', with: 'password'
+    fill_in 'Password', with: 'test1234'
+    fill_in 'Password confirmation', with: 'test1234'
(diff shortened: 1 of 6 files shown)


spec/features/update_password_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  let(:person) { create(:person, password: 'password') }
+
+  scenario 'update password' do
+    visit person_path(person, redesign: true)
+    click_on 'Change password'
+    fill_in 'Current password', with: 'password'
+    fill_in 'Password', with: 'test1234'
+    fill_in 'Password confirmation', with: 'test1234'
+    click_on 'Update'
+    expect(page).to have_content 'You updated your account successfully'
```
til: Maybe we also want to assert that the password was really changed to "test1234".

spec/features/update_password_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+
+feature 'update password' do
+
+  before { visit new_person_session_path }
+  before { sign_in person }
```
til: It doesn't make sense to have separate before blocks in this case I think, rather

```
before do
  ...
end
```

(I know, copied from a bad precedence ;-))

til: Conceptual overhead!

til:
Cool :rocket: :cocktail:

PR #492, opened 2016-11-24
#483: Adjust views in app/view/devise/registrations to change password

katrin-k:
Solves issue #483, regenerating the required view to change a user's password. 

(Note: The submit-button is currently not displayed correctly. Will be solved by the next PR for the redesign. Coming in 3…2… today…)

### DIFF
--- a/app/views/devise/registrations/edit.html.haml
+++ b/app/views/devise/registrations/edit.html.haml
@@ -1,0 +1,31 @@
+%h2
+  Edit Password of #{resource_name.to_s.humanize}
+= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
+  = devise_error_messages!
+  .field
+    = f.label :email
+    %br/
+    = f.email_field :email, autofocus: true
+  - if devise_mapping.confirmable? && resource.pending_reconfirmation?
+    %div
+      Currently waiting confirmation for: #{resource.unconfirmed_email}
+  .field
+    = f.label :password
+    %i (leave blank if you don't want to change it)
+    %br/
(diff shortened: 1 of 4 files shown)


til:
Cool!

This changes some of the views for the current design - did you check
that the current registration still works as intended? Well, the tests
seem to pass, so that is a good sign, but I'm not too sure that they
cover it completely, and definetely they don't cover how it looks.

app/views/devise/registrations/edit.html.haml
```diff
@@ -0,0 +1,31 @@
+%h2
+  Edit Password of #{resource_name.to_s.humanize}
+= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
+  = devise_error_messages!
+  .field
+    = f.label :email
+    %br/
```
til: Just  a minor remark: the slash is not necessary, just `%br` is sufficient. It will produce a `<br>` tag, which is allowed (must be?) to be empty. It seems the generator is doing the wrong thing here.

katrin-k:
Right… Yes, I did double-check functionality by 1) changing the password and 2) signing up with a new user. 

Regarding the new views: They are obsolete within the redesign and in the old design, there's no link to changing the password. So I figure, that's fine so far. Sorry for not pointing this out earlier. 
(I guess, there will be quite some tidying up anyhow, once we launch.)

katrin-k:
@til And on this one: If you don't have any other thoughts/concerns, I'd also give it a merge.

til:
Fine for me.

til:
Sorry, took some time, now it's deployed.

Issue #488, opened 2016-11-15
Relaunch: must haves


katrin-k:
For the relaunch of the new design, I'd like to open this new issue where we can focus on the must-haves for the relaunch. It's an add-on for #468.
Feel free to reorder the list as deemed necessary. 

### Todos before relaunch:
- [x] Test the hell out of the page. 
- [x] Button pattern (somewhat consistent atm, but not really. Another button style might also be in order, especially to mark differences to tags.) (Katrin)
- [x] Check border on buttons that are `<input type="submit>`
- [x] All links need care. They currently don't match the design by Justine. (Katrin)
- [ ] Resposinvenes/ media queries
- [x] Check all files for "TODO"-comments: Here I mentioned when I had a problem somewhere 
- [ ] Rework welcome page
- [x] Alert messages
- [x] Profile page: display of the groups

### Todos during relaunch:
- [x] remove `appendRedesignParam()` from main.js 
- [x] remove all Bootstraps

### Todos right after relaunch:
- [ ] test the page for missing functionality (Bootstraps' JS might interfere beforehand)
- [ ] Tabbed content for groups#show

#### Nice-to-haves
The nice-to-have list has issues that are no deal-breakers and can be reassigned to separate issues later on. 
- [ ] Enhance form styling for error highlighting, focus states, etc.
- [ ] Check all files for missing accessibility markup
- [ ] Admin dashboard: functionality of the gif section

berlintam:
@katrin-k here are my notes!

1. the profile page is a bit squishy, & also clicking the change password button doesn't do anything. Also the "working on" section doesn't seem to respect markdown anymore. 
<img width="1002" alt="screen shot 2016-11-15 at 20 32 50" src="https://cloud.githubusercontent.com/assets/4237285/20320650/d3d6d6d2-ab72-11e6-8aff-da106fe55a7d.png">

2. groups page fails if the markdown content is nil (ex: if group.activities or group.activities is empty).
<img width="1268" alt="screen shot 2016-11-15 at 20 42 10" src="https://cloud.githubusercontent.com/assets/4237285/20321023/475529f0-ab74-11e6-970e-762a453854f5.png">

3. top part of welcome page looks like an error, instead of just information. Maybe just changing the colour will help?  & GAAASP Captain Picard eating cake is gone!
[...]

PR #493, opened 2016-11-24
Redesign: button, link, whitespace

katrin-k:
Following #488: 
- Consistent link and button patterns
- All of the whitespace/margin issues


til:
If you want me or someone to merge and deploy, let us know. It only
affects the redesign version, right? I only had time to look over it
briefly, looks good so far.

katrin-k:
@til I'm happy to merge, but I can't do the deployment part. If you have time for deployment.
And yes, this PR only affects the redesign.

til:
OK, I deployed now, and did the

[ror@schedar rorganize.it]$ chmod +x scripts/uptime.sh

until that is fixed, so that Lisa does not get lots of mail again.

In the future it would be cool if more people could deploy besides me,
so if you don't mind, let's set it up the next time we meet.

PR #494, opened 2016-11-29
make topics and comments visible to everyone, close #481

### LINKED ISSUE #481: Who can read topics
We never actively decided how this should work, so let's do it now.

**Old state:** topcis were listed on the main group page. Everyone could see it, members could administrate

**Current state:** topics live nested under groups, only members can acces then at all

**Wanted state:** ???

lislis:
but keep edit functionality excusive to group members

### DIFF
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -1,7 +1,6 @@
 class TopicsController < ApplicationController
   before_action :set_topic, only: [:show, :edit, :update, :destroy]
-  before_action :validate_user_group_member
-  before_action :authenticate_person!
+  before_action :authenticate_person!, only: [:edit, :update, :destroy]
 
   def edit
   end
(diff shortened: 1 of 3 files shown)


app/controllers/topics_controller.rb
```diff
@@ -1,7 +1,6 @@
 class TopicsController < ApplicationController
   before_action :set_topic, only: [:show, :edit, :update, :destroy]
-  before_action :validate_user_group_member
```
til: So if we don't validate a user group member anymore, does that mean anyone who is authenticated can edit the topics?

lislis: well, I can't access the topics page when I'm not a member. Does that count?

til: Hm, the controller code looks as if any logged in person, also non-members, can create, update and destroy any topic. Where is the group membership checked after that `before_action` is removed? Also, the private method `validate_user_group_member` is now unused I think.

lislis: hmmm, not sure. the code reads like it so from that point of view it makes sense to get it back. But I don't know why I didn't run into that problem locally

Issue #284, opened 2014-11-26
non-public/public profile


1000miles:
dear rubycorns, 

for privacy reasons i would love to suggest following feature:
- provide a button for every user to switch from "visible" (in public listing) to "hidden" (only visible for logged-in users)

what do you think?

best,
thuy

berlintam:
hey, what's the status on this? @1000miles did you still want to take a look at it?

1000miles:
@berlintam I am sorry, I seem to have missed your comment from 12 days ago. 
What do you mean by taking a look at it? As far as I can see the profile names are still public to everyone, so this task is still in backlog, or? Waffle says yes. :)

berlintam:
sorry @1000miles! Ignore the waffle. I was bored one day but it doesn't represent anything at this point. I'm just gonna delete it. Feel free to take a crack at this if you want, or not, everything works :)

til:
#497 is dealing with this.

lislis:
This is now live! Closing

PR #497, opened 2017-01-03
Add private profiles, closing #284

lislis:
I added a 'private' flag to user profiles.
When your account is private your account is not visible at all for non-registered/ non-signed-in users.
This means that it does not show on the people overview page or the individual group pages.

Todo: add tests

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -54,6 +54,11 @@ class GroupsController < ApplicationController
 
     @topic = Topic.new
     @topic.group = @group
+
+    @students = @group.students.public_profile
+    @students = @group.students if signed_in?
+    @coaches = @group.coaches.public_profile
+    @coaches = @group.coaches if signed_in?
   end
 
   def destroy
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -4,6 +4,7 @@ class PeopleController < ApplicationController
 
   def index
     @people = Person.order(:first_name).order(:last_name)
+    @people = @people.public_profile unless signed_in?
     @people = @people.by_country(params[:country]) if params[:country].present?
     @people = @people.by_city(params[:city]) if params[:city].present?
 
@@ -34,6 +35,6 @@ class PeopleController < ApplicationController
 
   def person_params
     params.require(:person).permit(:first_name, :last_name, :email,
-    :picture, :twitter, :website, :working_on, :workshop_coach, :city, :country, :willing_to_travel)
+    :picture, :twitter, :website, :working_on, :workshop_coach, :city, :country, :willing_to_travel, :non_public)
   end
 end
--- a/app/views/groups/show.html.haml
+++ b/app/views/groups/show.html.haml
@@ -95,18 +95,24 @@
 
       %h4 Students
       %p
-        This group currently has #{pluralize @group.students.length, "student"}.
+      - if signed_in?
+        This group currently has #{pluralize @students.length, "student"}.
+      - else
+        This group may have more students than currently shown. Log in to see all of them.
       %ul.list--people
-        - @group.students.each do |student|
+        - @students.each do |student|
           %li
             = link_to student.first_name, person_path(student)
             - if admin?
               = render partial: 'remove_from_group', locals: { object: student, group: @group }
       %h4 Coaches
       %p
-        This group currently has #{pluralize @group.coaches.length, "coach"}.
+      - if signed_in?
+        This group currently has #{pluralize @coaches.length, "coach"}.
+      - else
+        This group may have more coaches than currently shown. Log in to see all of them.
       %ul.list--people
-        - @group.coaches.each do |coach|
+        - @coaches.each do |coach|
           %li
             = link_to coach.first_name, person_path(coach)
             - if admin?
--- a/app/views/people/_form.html.haml
+++ b/app/views/people/_form.html.haml
@@ -22,6 +22,10 @@
       .form-group
         = f.label(:working_on, "What are you working on? (you can use markdown!)")
         = f.text_area :working_on, placeholder: 'tutorials? a cool app?', class: 'form-control', rows: 5, id: 'working-on'
+      .form-group
+        = f.check_box(:non_public)
+        = f.label(:non_public, "Make my account not public")
+        %p.help-block Your account will only be visible to other registered users
 
     .col-md-1
 
@@ -32,21 +36,21 @@
       %ul
         %li You may be contacted by Rails Girls organizers about coaching at various workshops
         %li We will share your email address with trusted organizers
-        %li 
-          Your name will be displayed on our 
-          %a{ href: coaches_path } coaches index page. 
+        %li
+          Your name will be displayed on our
+          %a{ href: coaches_path } coaches index page.
         %li Everyone will think you're really really awesome. And most importantly, you will be.
       %br
       %p Convinced? That's great. We just need a few things from you,
 
       .form-group
         = f.check_box(:workshop_coach)
-        = f.label(:person_workshop_coach, "I am willing to COACH at Rails Girls workshops")
+        = f.label(:workshop_coach, "I am willing to COACH at Rails Girls workshops")
 
       .form-group
         = f.check_box(:willing_to_travel)
         = f.label(:willing_to_travel, "I am willing to TRAVEL to Rails Girls workshops")
-      
+
       .form-group
         = f.label(:city, "City")
         = f.text_field :city, placeholder: 'Rubycity', class: 'form-control'
--- a/config/autoprefixer.yml
+++ b/config/autoprefixer.yml
@@ -1,0 +1,3 @@
+browsers:
+  - "last 2 versions"
+  - "IE > 10"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160621202750) do
+ActiveRecord::Schema.define(version: 20170127155731) do
 
   create_table "comments", force: :cascade do |t|
     t.string   "person_id",  null: false
@@ -101,6 +101,7 @@ ActiveRecord::Schema.define(version: 20160621202750) do
     t.string   "city",                   limit: 255
     t.string   "country",                limit: 255
     t.string   "website",                limit: 255
+    t.boolean  "non_public",                         default: false, null: false
   end
 
   add_index "people", ["email"], name: "index_people_on_email", unique: true
--- a/app/controllers/coaches_controller.rb
+++ b/app/controllers/coaches_controller.rb
@@ -1,6 +1,7 @@
 class CoachesController < ApplicationController
   def index
     @coaches = Person.workshop_coach.order_by_name
+    @coaches = @coaches.public_profile unless signed_in?
     @coaches = @coaches.by_country(params[:country]) if params[:country].present?
     @coaches = @coaches.by_city(params[:city]) if params[:city].present?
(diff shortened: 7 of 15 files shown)


PragTob:
Tests fail :( Looks like a thing where we don't have a current_user or something in the feature specs

app/controllers/groups_controller.rb
```diff
@@ -54,6 +54,17 @@ def show
...
+    @students = @group.students.select do |x|
+      if signed_in? || !x.private?
+        x
+      end
+    end
+    @coaches = @group.coaches.select do |x|
+      if signed_in? || !x.private?
+        x
+      end
+    end
```
PragTob: I think we can improve here! :)

I think ideally we'd do a `public` scope on people so that if we only want to see the public ones we could do:

`@students = @group.students.public`

the scope would be something like `scope(:public) { where(private: false) }` (might be slightly different I haven't done Rails in ~3 months ;P)

then the controller could be simplified to:

```ruby
if signed_in?
  @students = @group.students
  @coaches = @group.coaches
else
  @students = @group.students.public
  @coaches = @group.coaches.public
end
```

We could also go more fancy... e.g. the authorization I commonly use defines authorization in terms of scopes as access to objects that would then define the access to students/coaches globally and here we'd just do something like `@students = power(:students)` but that might be getting too fancy. If you're interested though, the library is [consul](https://github.com/makandra/consul) :)

app/controllers/people_controller.rb
```diff
@@ -4,6 +4,7 @@ class PeopleController < ApplicationController
 
   def index
     @people = Person.order(:first_name).order(:last_name)
+    @people = @people.reject {|x| x.private?} unless signed_in?
```
PragTob: we could use the scope here again and would avoid repeating the concept encapsulated in the reject block! :)

app/views/groups/show.html.haml
```diff
@@ -106,11 +106,12 @@
       %p
         This group currently has #{pluralize @group.coaches.length, "coach"}.
       %ul.list--people
-        - @group.coaches.each do |coach|
+        - @coaches.each do |coach|
           %li
             = link_to coach.first_name, person_path(coach)
             - if admin?
               = render partial: 'remove_from_group', locals: { object: coach, group: @group }
+      %p Please note that not all profiles are shown publicly.
```
PragTob: Not sure if we need to tell people this... maybe we do. Not sure :)

config/autoprefixer.yml
```diff
@@ -0,0 +1,3 @@
+browsers:
+  - "last 2 versions"
+  - "IE > 10"
```
PragTob: :+1:

db/migrate/20161129232712_add_public_profile_to_person.rb
```diff
@@ -0,0 +1,5 @@
+class AddPublicProfileToPerson < ActiveRecord::Migration
+  def change
+    add_column :people, :profile_visibility, :string, :default => ''
+  end
+end
```
PragTob: I don't think we need this migration or do we?

db/schema.rb
```diff
@@ -101,6 +101,8 @@
     t.string   "city",                   limit: 255
     t.string   "country",                limit: 255
     t.string   "website",                limit: 255
+    t.boolean  "public_profile"
```
PragTob: I don't think we need this entry or do we? :)

app/views/people/_form.html.haml
```diff
@@ -22,6 +22,10 @@
       .form-group
         = f.label(:working_on, "What are you working on? (you can use markdown!)")
         = f.text_area :working_on, placeholder: 'tutorials? a cool app?', class: 'form-control', rows: 5, id: 'working-on'
+      .form-group
+        = f.check_box(:private)
+        = label_tag(:private, "Make my account private")
```
til: Would it be possible to use `f.label` instead here? (Just for consistency with the rest of the file)

lislis:
I tried to address the points you raised. There is still the `authenticate?` error and I think my tests (besides from failing themselves) also introduce a seemingly unrelated error?

til:
Thanks for the fixes, looks good except for the failing controller
specs.

I think I have found the reason for it. The error message says

       undefined method `authenticate?' for nil:NilClass
       # /home/tils/.gem/ruby/2.3.0/gems/devise-3.5.6/lib/devise/controllers/sign_in_out.rb:10:in `block in signed_in?'


that line within the devise source code is this:

       warden.authenticate?(scope: _scope)

so warden, which is something used by devise, is nil. Further down in
the error message the backtrace includes

     # ./app/controllers/people_controller.rb:7:in `index'

which has the call to `signed_in?`. Changing the line to simply calling
`signed_in?` in the controller yields the same error, commenting it out
fixes the spec, so it definitely is something with devise.

On the devise README at https://github.com/plataformatec/devise there is
a note on testing controllers which seems to apply here. However, the
latest version of devise is a lot newer than the one we currently use.
So one needs to look at the README of our bundled devise gem. Thanks to
`bundle open` this is quite easy.

I'll leave the final fix to you or someone else who wants to grab it ;)

lislis:
thanks @til! looking at https://github.com/plataformatec/devise/blob/v3.5.6/README.md#test-helpers and adding one line fixed things!

db/schema.rb
```diff
@@ -101,6 +101,7 @@
     t.string   "city",                   limit: 255
     t.string   "country",                limit: 255
     t.string   "website",                limit: 255
+    t.boolean  "private",                            default: false
```
raluka: what do you think about add condition `null: false`, just to be sure? 
Till pointed out that `nil == false` returns `false` 💡

app/views/people/_form.html.haml
```diff
@@ -22,6 +22,10 @@
       .form-group
         = f.label(:working_on, "What are you working on? (you can use markdown!)")
         = f.text_area :working_on, placeholder: 'tutorials? a cool app?', class: 'form-control', rows: 5, id: 'working-on'
+      .form-group
+        = f.check_box(:private)
+        = f.label(:private, "Make my account private")
+        %p.help-block Your account will only be visible to other registered users
```
raluka: From what I read in PR description, we don't display private profiles to any kind of users, registered or not. What about admins? Can they see a private profile? Maybe we can mention this in help-block

config/autoprefixer.yml
raluka: I think we have a missing line at EOF

spec/features/people_private_profile_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'spec_helper'
+
+feature 'person profle visibility' do
```
raluka: I am big fan of tests, so I propose to add 2 more tests: to check that private profile is not seen by a logged in user, and to test when logged in as admin user.

app/controllers/groups_controller.rb
```diff
@@ -54,6 +54,9 @@ def show
 
     @topic = Topic.new
     @topic.group = @group
+
+    @students = @group.students.public_profile
```
raluka: I'll also add here check for `signed_in?`

app/views/groups/show.html.haml
```diff
@@ -97,7 +97,7 @@
       %p
         This group currently has #{pluralize @group.students.length, "student"}.
```
raluka: What about "This group composition has private settings. You can find more informations after login" - or something similar, in case one (or more, or all) members have private profiles? This instead of showing count for members

til: I agree that this makes sense - showing a number when the number of displayed names does not correspond would be irritating. We started implementing something for this in a helper, but then got sidetracked into a yak shave with the seeds. So maybe next time more on this.

db/schema.rb
til: 👍

lislis:
@til I just realized that the migration seems to have a weird date 0.o would this even work on our production db?

til:
Lol, yeah, that date looks sort of old. But as far as I know it should
work nevertheless - the db:migrate task will run all migrations that
have not been applied to the database yet, so it will run this one. The
dates are only useful for ordering and avoiding conflicts between
different branches adding migrations in parallel (they used to be
increasing numbers, which resulted in such conflicts).

For cosmetic reasons, you could rename the migration to a more recent
date. You'll need to switch to master, reset your database, switch back
to this branch, then do the rename and then run migrations, otherwise it
would try to run it twice in your local database.

lislis:
woof. I think this is it?

app/views/groups/show.html.haml
```diff
@@ -94,19 +94,27 @@
...
 
       %h4 Students
-      %p
-        This group currently has #{pluralize @group.students.length, "student"}.
+      - if signed_in?
+        %p
+          This group currently has #{pluralize @group.students.length, "student"}.
+      - else
+        %p
+          This group may have more students than curently shown. Log in to see all of them.
```
til: currently

til: Since the `%p` is the same in both cases, it could be on the outside of the if statement, then it wouldn't need to be repeated.

Below, the new instance variable `@students` is used to loop over the students to display, which is good. I think it would be consistent to use the same instance variable to display the count as well in the sentence above - otherwise someone might misunderstand them as being deliberately different, which they are not (or are they?).

app/controllers/groups_controller.rb
```diff
@@ -54,6 +54,14 @@ def show
...
     @topic = Topic.new
     @topic.group = @group
+
+    if signed_in?
+      @students = @group.students
+      @coaches = @group.coaches
+    else
+      @students = @group.students.public_profile
+      @coaches = @group.coaches.public_profile
+    end
```
til: This is already good, but I nevertheless have a remark with a suggested improvement: in cases like this, it is important to make sure that in all branches of the if clause the same variables are assigned. While it is hard to miss in this case (e.g. it would be unlikely to forget to assign `@coaches` in the else branch and not notice it immediately), it might still be nice to change it so that this cannot happen, and also remove some repetition. In ruby one can assign to multiple variables at once, so the instance variables could be moved out of the if block:

```ruby
@students, @coaches = if
  ...
end
```

hmm, now that I think of it, there is an even more concise way. The way it has been done already in the people controller, maybe worth to copy that.

lislis:
@til hmmm, I get your point about the wording for private. I think I'll go for `public` then

lislis:
I renamed everything form 'private' to 'non_public'.
llama

til:
Thanks for putting all the work into this, looks very good already.

One thing that @raluka and I last time thought of when looking at this: are really all cases covered, or are there maybe some pages that leak non-public profiles to non-registered visitors? I found that CoachesController does not seem to check for non_public, so coaches that have the flag set would still be listed there. Maybe someone else wants to look through all controllers and see if there are more cases.

til:
👍 

Maybe a final squash and rebase? I think many of the commits here are actually from master?

lislis:
oh my. I'll fix this at home or maybe tomorrow

lislis:
dance bird  
I think this is it

app/controllers/groups_controller.rb
```diff
@@ -54,6 +54,11 @@ def show
 
     @topic = Topic.new
     @topic.group = @group
+
+    @students = @group.students.public_profile
+    @students = @group.students if signed_in?
+    @coaches = @group.coaches.public_profile
+    @coaches = @group.coaches if signed_in?
```
til: This here is done slightly different than in the other controllers - here it completely replaces the previous collection, whereas in the other controllers it refines it. For consistency this could be refactored, but I'm going to merge this now, so if you want to do so in a new PR.

til:
🚀 ... will someone else deploy please?

lislis:
Deploying and opening a refactor ticket. Thanks @til!

lislis:
already up and running!

PR #499, opened 2017-01-14
only highlight logged in user on coach page (fixes #498)

### LINKED ISSUE #498: coaches page thinks everyone is me!
If you go to the coaches index page, behaviour is a bit odd. All users are highlighted (background is yellow), and the text beside them says: "it's you". Strange thing is, I'm not logged in. 

Ideally, this behaviour should only happen for the logged in user, not everyone. See screenshot for clarification.

screen shot 2017-01-14 at 11 28 58

update: it seems as though this only happens when the redesign params are there.

berlintam:
@katrin-k I'm not sure if you already noticed/fixed this. if so, feel free to delete this PR. If not, feel free to merge (provided you're happy with this & it makes sense). 

Lemmie know if you have any questions/concerns.

### DIFF
--- a/app/views/coaches/index.html+redesign.haml
+++ b/app/views/coaches/index.html+redesign.haml
@@ -23,7 +23,7 @@
     %hr
     %ul.list-group
       - @coaches.each do |coach|
-        %li{class: "list-group-item #{'list-group-item-warning' if logged_in?(current_person)}"}
+        %li{class: "list-group-item #{'list-group-item-warning' if logged_in?(coach)}"}
           .list-group-item-pic
             = person_avatar(coach)
           .list-group-item-info
@@ -31,7 +31,7 @@
               = profile_link(coach)
             - if logged_in?(current_person) && coach_location?(coach)
               %p #{coach.city}, #{coach.country_name}
-          - if logged_in?(current_person)
+          - if logged_in?(coach)
             .list-group-item-person
               It's you
 


PragTob:
Oh, wait I approved and it's all good - but especially for regressions I feel writing a test covering this so that it doesn't reoccur is super important. So if someone wants to tackle that, I'd be super :tada:

berlintam:
@PragTob you mean a feature spec for this page? or what part specifically?

While I was fixing this I noticed that we have no specs for the people_helper, so I'm in the process of doing that now. Is that what you mean?

PragTob:
Whatever spec could have caught that bug :)

The way it is implemented right now, only a feature spec for this spec could have caught it imo. So that'd be the rout to go. However, one could decide that this would be slightly overkill - and extract it into a helper method... but then again that probably wouldn't have called it as it was more of a thing of passing the wrong parameter rather than a business logic error in the method.

berlintam:
I totes agree. 

But.... 😄  This was on the re-design page, & I'm not sure how we are going forward with specs for this. Ideally I would write a feature spec for visiting the page after agreeing to coach at workshops & making sure my name is there & that it's me etc... But from what I can see, all feature specs are testing non redesign pages. We should probably handle this somehow......

PragTob:
Argh you're right... then.. ermm.. I dunno. Ideally the redesign wouldn't have new features so we could spec against the old page and it would still work with the new page? That's wishful thinking though... maybe we could run it against the redesign and once we make it the main just remove the flag? Or we create an issue to write the spec once the redesign is main... dunno :o

berlintam:
I dunno either. Maybe we can take a vote on Tuesday/come up with a super cool idea that handles all possible edge cases & simultaneously supplies us all with chocolate and 🍰

PR #500, opened 2017-01-15
refactor & test people helper methods

berlintam:
I was looking thru the people_helper, & noticed we are using both `person_signed_in?` and `logged_in?(person)` . They both looked like the same method to me, & after some digging, I saw that `person_signed_in?` comes from [devise](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/controllers/helpers.rb#L52). 

We were also using the two methods interchangeably, we were calling `logged_in?(current_person)` and also `person_signed_in?`, when they both (I believe) return the same thing. 

So I figured I'd get rid of `person_signed_in?` and just use our own method instead (`logged_in?`). It's also a bit more dynamic, as it can compare two people (ex: is the current person this other person I'm looking at? If so, I can edit their page) by passing in an argument, or without an argument, it just returns true or false if the current_person is logged in (default arguments ftw!)

Along the way, I also noticed we were missing a chunk of specs, so I added those. Oh, and I renamed the `coach_location?` method because non-coaches should in theory also be able to have locations. 

review/opinions appreciated :)


app/helpers/people_helper.rb
```diff
@@ -1,32 +1,31 @@
 module PeopleHelper
-  def logged_in?(person)
-    current_person && current_person == person
+  def logged_in?(person = current_person)
```
sareg0: why are you passing are you making the default `person = current person`? I'm just curious

sareg0: Oh I see!! Unless we pass someone else other than `current_person` in, we'll always just check the current person.  I understood by looking at the view.

app/views/pages/welcome.html.haml
```diff
@@ -93,7 +93,7 @@
                   = markdown(":cake:")
 
           = link_to "Go To Groups!", groups_path, class: 'btn btn-custom'
-          
+
```
sareg0: it looks like there are a few empty lines at the end of this file for no reason. Could I trouble you to remove them, just to aid in cleanup?

spec/helpers/people_helper_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'spec_helper'
+
+describe PeopleHelper do
+  let(:null_person) { NullPerson.new }
+  let(:person) { build_stubbed :person }
+
+  describe '#logged_in?(person)' do
+    before { allow(helper).to receive(:current_person).and_return person }
+
+    specify 'checks if current person is logged in by default' do
```
sareg0: maybe just go through and unify the use of double quotes throughout the file

sareg0:
I made some minor comments regarding the unification of using single or double quotes, and also about some extra space we can get rid of, but once those things are done, I would be happy to merge this

sareg0:
Actually one more thing. i noticed that in the spec files you are using `specify` in your specs. For consistency, could you use 'it' for your specs?

I read in the [better specs docs](http://betterspecs.org/) that it is good to write the string afterward with a doing verb, like so

```
it 'responds with 200'
it 'shows the resource'
```

berlintam:
Thanks @sareg0!!!
I'm happy to change `specify` to `it`. We use `specify` at work, so it's totes habit. Thanks for linking to better specs! Haven't looked at it in a while 😱

berlintam:
@sareg0 do you think this is ok to merge?

sareg0:
I'm gonna merge it in now!! Hooray!

sareg0:
!()[https://media.giphy.com/media/3oz8xRF0v9WMAUVLNK/giphy.gif]

PR #501, opened 2017-01-15
allow users to sign up again

berlintam:
looks like our current sign up behaviour is broken. A user is required to provide us with a first name when signing up, but our form only asks for an email & password. So try as you might, you can't sign up!
https://rorganize.it/people/sign_up

This fixes it. We also didn't have a spec for signing up, so it's understandable that this slipped through the cracks.


PragTob:
Also: :scream: 

Anyone deploycorning?

PR #505, opened 2017-02-04
refactor groups controller collections

### LINKED ISSUE #504: Refactor groups controller
screen shot 2017-02-02 at 16 18 12

berlintam:
Closes #504 

@lislis & @til is this what you meant?

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -55,10 +55,8 @@ class GroupsController < ApplicationController
     @topic = Topic.new
     @topic.group = @group
 
-    @students = @group.students.public_profile
-    @students = @group.students if signed_in?
-    @coaches = @group.coaches.public_profile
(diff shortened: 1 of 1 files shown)


PragTob:
:+1:

til:
> @lislis & @til is this what you meant?

My main concern with the code is fixed. Previously, `@students` was
assigned to something and in the next line, that assignment was
completely discarded again if `signed_in?` was true. I can't explain
exactly why but it didn't feel right. Now `@students` is only assigned
once, so that's good (and the same for `@coaches`).

In
https://github.com/rubycorns/rorganize.it/blob/master/app/controllers/people_controller.rb#L6-L7
it's done a bit differently - the initial assignment is modified in the
lines further down, based on the conditions. But I guess that's mainly
inspired from how later on there the filters are applied, either way is
cool.

PR #506, opened 2017-02-06
Refactor filtering in controllers

berlintam:
Issue #504 from @til & @lislis made me take a closer look at our controllers & how we handle our filtering and I thought I'd take a crack at refactoring it. I created the `filtered_by_params` class method which can now be used across all controllers to filter the cities, countries, & willingness to travel for groups & people & also takes into account public/non public profiles. 

I also noticed that we were showing all possible cities/countries in our drop downs, even if that city/country belonged to someone who has a non_public profile. I created the `visible_locations_for` scope to handle this case which is also now used in the controller.

Also, I used a switch statement. Which like... I dunno. I'm not a huge fan of switch statements for some reason, but I thought it fit the use case.....

Review is very much appreciated & if anyone has any questions let me know!


lislis:
Such refactorings! Wow

app/controllers/groups_controller.rb
```diff
@@ -7,15 +7,11 @@ class GroupsController < ApplicationController
...
-    @groups = @groups.by_country(params[:country]) if params[:country].present?
-    @groups = @groups.by_city(params[:city]) if params[:city].present?
-
-    @groups_inactive = @groups_inactive.by_country(params[:country]) if params[:country].present?
-    @groups_inactive = @groups_inactive.by_city(params[:city]) if params[:city].present?
+    ordered_groups = Group.where(inactive: false).order(:name)
+    ordered_groups_inactive = Group.where(inactive: true).order(:name)
 
+    @groups = ordered_groups.filtered_by_params(params)
+    @groups_inactive = ordered_groups_inactive.filtered_by_params(params)
```
sareg0: while I like the symmetry of `@groups` and `@groups_inactive`, I feel like the name `@inactive_groups` reads better.

berlintam: totally agree. I actually changed it once and then I was like UUUUUUGH but then I'll have to change it everywhere so I changed it back. But now that you've also pointed it out I think it's worth changing.

app/models/concerns/search_filter.rb
```diff
@@ -0,0 +1,31 @@
...
+    scope :by_country, -> (country) { where(country: country) }
+
+    def self.filtered_by_params(params, signed_in = nil)
+      collection = (signed_in || self.name != 'Person') ? all : all.public_profile
+      case
+      when params[:country].present?
+        collection.by_country(params[:country])
+      when params[:city].present?
+        collection.by_city(params[:city])
+      when params[:willing_to_travel] == '1'
```
PragTob: While I like the notion of one common module for all, this here takes care of all the parameters - e.g. willing_to_travel - while this totally doesn't apply to groups for instance. And having that there for a module included in Group seems confusing to me. Not sure what to do about it. Maybe each class still implements there own as it's the responsibility of the model to see how it can be filtered?

I can see how there may be concerns of things not being DRY might pop up there. DRY however shouldn't be seen as much as pure code duplication but more as a duplication of concepts - which is arguable here as well. One could argue that the duplication here is "incidental" in that by coincedence both people and groups an be filtered by city/country whatever but there is no inherent deeper connection so that how they are filtered is two different concepts and therefore duplicating some of the code is ok. Or we could say that it inherent in the concept and what changes for one also changes for the other, as they are both a "Locatable" or whatever and then we should avoid the duplication.

I hope I'm making some sense, happy to chat about it!

spec/models/concerns/search_filter_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      it 'filters for German groups' do
+        params = { country: 'DE' }
+        expect(Group.filtered_by_params(params)).to match_array german_group
+      end
+
+      it 'filters for Torontonian groups' do
+        params = { city: 'Toronto' }
+        expect(Group.filtered_by_params(params)).to match_array canadian_group
+      end
```
PragTob: I'm missing tests for multiple parameters as I suspect it might not work and it's important as well :)

berlintam:
thanks Tobz!

app/models/concerns/search_filter.rb
berlintam: Doesn't seem to, should it? I've googled & I can't find anything that says it should. 

I can write:
`= check_box_tag 'willing_to_travel', true` [here](https://github.com/rubycorns/rorganize.it/blob/master/app/views/shared/_filter.html.haml#L17) which will send it as a boolean value instead.... but that strikes me as being a bit unreadable. 

Maybe I'm missing something? wdyt?

sareg0: @berlintam if you wanted to you could make the 'willing to travel' checkbox you linked to checked or unchecked depending on the params. I see that currently it becomes unchecked as soon as it is sent, so it is hard to follow what filters are still applied, at least visually and without paying attention to the params in the url. 

One can achieve this with an instance variable, and passing a `checked` option to the `check_box_tag`
`check_box_tag 'willing_to_travel', {checked: @willing_to_travel}`

`@willing_to_travel = params[:willing_to_travel] == "1"`

Hopefully I am correct in thinking this would assign `true` to `@willing_to_travel` if the params has a `1` for `willing_to_travel`

Maybe you want a different naming, or some such thing.

app/models/concerns/location_filter.rb
```diff
@@ -7,10 +7,18 @@ module LocationFilter
 
     scope :countries, -> { pluck(:country).uniq.compact.reject(&:blank?) }
     scope :by_country, -> (country) { where(country: country) }
+
+    def self.filtered_by_region(params, signed_in = nil)
+      collection = (signed_in || !self.respond_to?(:public_profile)) ? all : all.public_profile
+      if params[:country] || params[:city]
+        collection = collection.by_country(params[:country]) if params[:country].present?
+        collection = collection.by_city(params[:city]) if params[:city].present?
```
sareg0: I am unsure of how this works if there are both country and city params present. Is this even a possibility? 

To me it looks like `collection` is assigned in the event there are `params[:country].present?` and then collection could end up being assigned again if `params[:city]` is present. I feel like perhaps this is an 'elsif' in disguise.

app/models/concerns/location_filter.rb
```diff
@@ -7,10 +7,18 @@ module LocationFilter
...
+      end
+      collection
+    end
   end
 
   def country_name
     ISO3166::Country[country].name if country.present?
   end
-
 end
```
sareg0: I presume that little red symbol means that the file needs one empty line at the end, yes?

til: Yes, every well behaved text file should have one newline character (a "\n") at the end, for aesthetics and consistency (and to make diffs nicer when adding new lines after the end later). If it is missing, then GitHub adds this nasty red symbol to point that out. Some text editors are configured to automatically add one if it is missing.

app/models/concerns/location_filter.rb
til: You are correct, it is a possibility, and you are correct that the assignment occurs twice. Is that the right thing to do? To answer the question, we need to understand with what exactly the assignment occurs in the first and then in the second step. Is the assignment from the first step discarded when the second step happens? That would point out that it might be a bug.

In any case, it would be great to have a spec (maybe a high level feature spec) that exercises the case where both are set. I don't think we have that yet, or do we?

til: Never mind, there is this spec which tests what happens when both params are set: https://github.com/rubycorns/rorganize.it/pull/506/files#diff-885201b6b2d0fe6b146005b0db8001b5R22

berlintam: is this ok then? Or can someone suggest a better way to write it?

app/models/concerns/search_filter.rb
berlintam: ok, so what I have is:
`= check_box_tag 'willing_to_travel', 1, @willing_to_travel` in the view (the 1 is needed because I need to pass in a value to the check box tag before I can pass in whether or not the box can be checked, if I omit it, it passes in @willing_to_travel as a value for the checkbox which hasn't been defined yet and will therefore always be false)

Then in the controller I have: `@willing_to_travel = params[:willing_to_travel] == '1'`

This makes sure the checkbox is still checked after the filtering request has gone thu (if the checkbox was checked in the first place). In the filter, `@willing_to_travel` is now true, as you said. 

But, It feels a bit messy to me, because I'm passing in `1` in the checkbox tag(I can also pass in `true` but it goes thu the params as a string & then in the controller I have to to `== 'true'` which is just ugh!) It also doesn't solve the issue that Tobi brought up, which tbh I don't know if rails actually does & I can't find any info on it. Maybe the converting 1 to true behaviour happens with a gem like simple_form?

However, all that said, I do like the new behaviour or having the checkbox still tagged after the request has been made & is showing the filtered response. It's much clearer. So that's super cool & I'm totes willing to push these changes for that to happen. 

phew. long msg. thoughts?

app/models/concerns/location_filter.rb
```diff
@@ -10,11 +10,15 @@ module LocationFilter
...
-        collection = collection.by_country(params[:country]) if params[:country].present?
-        collection = collection.by_city(params[:city]) if params[:city].present?
+      if params[:country] && params[:city]
+        collection.by_country(params[:country]).by_city(params[:city])
+      elsif params[:country]
+        collection.by_country(params[:country])
+      elsif params[:city]
+        collection.by_city(params[:city])
+      else
+        collection
```
berlintam: @sareg0 & @til I rewrote this section using an if/else for clarity. Does it make more sense now? (I totes think it does)

app/models/concerns/search_filter.rb
til: Whether Rails converts the willing_to_travel parameter to a boolean is an interesting question. As far as I can see it does not - this is because Rails does have no idea that it should be a boolean! At the point where it has to decide that, it only gets a string with all the parameter values, e.g. like in this URL String from a GET request to the page - everything on the right of the question mark are the parameters:
https://rorganize.it/coaches?utf8=%E2%9C%93&city=&country=&willing_to_travel=1

Rails parses that, and only knows there is a param with the name "willing_to_travel" and the value is "1".

The conversion would happen if Rails would assign the parameters to a model and that model had a boolean attribute - at that point Rails knows that it has to convert the "1" String to a boolean `true` or `false`.

I think the way you are doing it is necessary and totally ok - convert the param value manually to a boolean. It should be done as early as possible, so that as soon as possible the code deals with a more meaningful set of filter criteria, converted to meaningful datatypes. The fact that the `willing_to_travel` attribute comes from an html form is a detail that is not interesting anymore when dealing with the filtering functionality.

til: Oh and by the way, it's helpful to recognize that `check_box_tag` is only marginally involved in what is happening here - in the end it just produces html code for the input tag, which is displayed in the browser. Theoretically that html could have been created statically or created with php as well.

spec/models/concerns/location_filter_spec.rb
```diff
@@ -1,6 +1,56 @@
 require 'spec_helper'
 
-RSpec.describe LocationFilter do
+RSpec.describe LocationFilter, :vcr => {:cassette_name => "create_group" } do
```
til: Why does this need a vcr cassette? Are there any external http requests happening when calling the LocationFilter?

(side note: use modern hash syntax)

app/models/concerns/search_filter.rb
til: I agree - dealing with both groups and coaches filtering within the same class doesn't fit. They are two concepts that do have some things in common, and also some specialities that they do not have in common (which is just a fancy way of saying, groups also has the attribute `willing_to_travel`).

Hmm, could it be that there is a pattern in object oriented programming for this kind of sharing common behaviour and separating special behaviour? 😄

berlintam: @til I've already refactored this out & moved `willing_to_travel`https://github.com/rubycorns/rorganize.it/pull/506/commits/7a04d68336ad3fb2d62281504feec51172d01af3 . Is that what you mean?

spec/models/concerns/location_filter_spec.rb
berlintam: looks like I do because I'm creating some groups further down in the file. When I remove it I get a vcr error
```
       An HTTP request has been made that VCR does not know how to handle:
         GET http://maps.googleapis.com/maps/api/geocode/json?address=%20%20%20Berlin%20DE&language=en&sensor=false
```
I haven't looked into it, but I can make an issue to investigate it further. I guess we shouldn't need a cassette every time we create a group.

app/models/concerns/location_filter.rb
til: To be honest, I don't think the current solution with the if/elsif blocks is ideal. There are four different possibilities already (none set, only city set, only country set, both set). Imagine what would happen if there was yet another attribute to be added - then there would be 8 options. And even within the code for the current 4 possibilities, a substantial amount of code is repeated.

Previously, the approach taken was like this:

1. first, assign the full collection to a variable
2. if necessary, reassign the variable limited by country
3. if necessary, reassign the variable limited by city

this way, no code needs to be duplicated, and it would be extendable with more attributes by adding new steps.

spec/models/concerns/location_filter_spec.rb
til: Ha, that's bad, it is trying to geocode the address of the new group. But not your fault and probably out of scope for this pull request, so maybe let's just keep it in mind and fix it later. 

Creating a group should not automatically trigger an external network request. This makes it very hard to create groups like here for the tests, or on the console or another place. The solution in my opinion is to make the call to geocode explicitly from the controller action where we really want it to happen.

app/models/concerns/location_filter.rb
berlintam: @til are you refering to the current state of the code or this code here? This is outdated & the new state of things is here: https://github.com/rubycorns/rorganize.it/pull/506/commits/7a04d68336ad3fb2d62281504feec51172d01af3. 

Do you think that the way it was previously done is more flexible then? I was seeing a lot of repetition across all controllers & was thinking if it was pulled out here, it would be easier to handle.

 If another attribute were to be added, it could still be added in the controller (like how now willing_to_travel is just in the coaches controller now and no longer handled here). This section is just filtering the city & country (location specific things). Perhaps more location things would be added, yes, like maybe..... kiez, but I can't really imagine that happening. 

>  And even within the code for the current 4 possibilities, a substantial amount of code is repeated.

What do you mean here? What part of it is repeated?

til: I was referring to the new state of things at https://github.com/rubycorns/rorganize.it/pull/506/commits/7a04d68336ad3fb2d62281504feec51172d01af3 where it is implemented as a large `if` statement. The approach to apply city and country filters to the collection how it was done previously was more flexible in my opinion, but I only meant the approach (what I outlined with the 3 numbered steps above), not the fact that this was repeated all over many controllers.

I think it's great that you are extracting and centralizing the filtering into a dedicated place, that is certainly a good thing.

And yes, it's unlikely that a third location filter attribute is going to be added - that was a very hypothetical point.

> > And even within the code for the current 4 possibilities, a substantial amount of code is repeated.
>
> What do you mean here? What part of it is repeated?

```ruby
      if params[:country] && params[:city]
        collection.by_country(params[:country]).by_city(params[:city])
      elsif params[:country]
        collection.by_country(params[:country])
      elsif params[:city]
        collection.by_city(params[:city])
      else
        collection
      end
```

The snippets `by_country(params[:country])` and `by_city(params[:city])` both appear twice. Not as obvious, but the conditionals checking for the presence of e.g. `params[:country]` are also somewhat repeated.

berlintam: ok, so perhaps then the way I did it before (in the commit we are currently commenting on) was more flexible? I just don't think it looks very nice & there's a lot of `if params[blahblahblah]` all over the place. Maybe it can just be refactored a bit?

I think though that the if/elsif/else statement is much more readable, so I'm leaning towards leaving it the way it is, but perhaps I can play around with refactoring it & removing some of the repetition. 

Thanks very much for your review @til!

learn

til: OK, referring to the commit that we are currently commenting on - If I understand, one outer wrapping if is not necessary. So instead of

```ruby
      collection = (signed_in || !self.respond_to?(:public_profile)) ? all : all.public_profile
      if params[:country] || params[:city]
        collection = collection.by_country(params[:country]) if params[:country].present?
        collection = collection.by_city(params[:city]) if params[:city].present?
      end
      collection
```

it could just be

```ruby
      collection = (signed_in || !self.respond_to?(:public_profile)) ? all : all.public_profile
      collection = collection.by_country(params[:country]) if params[:country].present?
      collection = collection.by_city(params[:city]) if params[:city].present?
      collection
```

Wouldn't that already be more readable?

When looking to refactor for readability, I'd put some effort into the first line, which looks a bit convoluted. What about extracting that into a method with a name that describes what it is doing, so that it could be called e.g. like this

```ruby
      collection = all_currently_visible(signed_in)
      collection = collection.by_country(params[:country]) if params[:country].present?
      collection = collection.by_city(params[:city]) if params[:city].present?
      collection
```

or maybe use the approach of re-assigning the collection based on some criteria for the first line as well (ok, it'll become two lines then):

```ruby
      collection = all
      collection = all.public_profile if limit_to_public_profile?(signed_in)
      collection = collection.by_country(params[:country]) if params[:country].present?
      collection = collection.by_city(params[:city]) if params[:city].present?
      collection
```

berlintam: looks cool, I'll give it a shot!

til:
I think it looks quite good already!

If I am allowed to further nitpick: one thing that makes me wonder now is, currently the `signed_in` boolean is passed in as second positional parameter to `filtered_by_region`, but what does filtering by public/non-public even have to do with a location filter? Those sound like different functionalities to me. Maybe that logic would be better put into its own place, where it is dealt with separately?

Then the call could look somehow like this:

```ruby
 @people = ordered_people
             .filtered_by_region(params)
             .filtered_by_visibility(signed_in?)
```

berlintam:
@til you're totally right. I changed it!
Thank you!!!!!

til:
Nice! From my point of view this PR looks good to merge now. Anybody
else wants to look at it? Or if you want Tam, go ahead and merge it.

Onwards to further refactorings!

berlintam:
giphy

Issue #370, opened 2015-08-25
paginate users page?


berlintam:
We now have 100 users on our page. This means we have the best app ever. 

Maybe we should think about paginating the people index. Or adding some search by alphabetisation (see all the users who's names start with A, for example). 

Or maybe just a search function. 

Wait a second... I'm getting deja vu....

lislis:
:confetti_ball: :tada: 100 users :tada: :confetti_ball: 
We could have a long list on one page and alphabetical jump links. Or a search of course. I like searches and then pagination.

berlintam:
by 100 I mean 145!

yes, alphabetical jump links is what I meant. much learn etc...

PragTob:
hm I never did alphabetical jump links, just pagination - both seem valid.

Search, of course, is great. However, if we stick with our "sqlite is valid in dev" we can't do the cool stuff as we can't use Postgres features. Then again, with 100 users we probably don't need fancy Postgres features, but Tobi LOVES fancy Postgres features.

til:
I love fancy postgres features as well! But I guess a list up to a
couple of hundreds of items would still be ok nowadays.

Grouping by alphabet is ok, but maybe there is an even better grouping
attribute that is more relevant to visitors - what about grouping by
city?

berlintam:
I love that idea!

But we don't really ask users for their location.... I mean... we do now sort of kind of via the coaching feature, but the average user isn't asked for their location.

til:
Unfortunately, yes. But maybe existing users will update their city
slowly, now that the feature exists?

To expand the grouping idea further - the groups index page could show
the cities as headlines and beneath the first x users of that city. When
the number of users in a city exceeds a certain threshold, it could have
a show more link:

Atlanta
  Anna A

Berlin
  Lisa P
  Tam E
  Til S
  <show more>

Copenhagen
  Smorreprot

etc. ...


This way, the visitors who are interested in members in a specific city
could find what they are looking for already on the groups index page.

berlintam:
I started working on this & I have a question: At the moment we only display people's location to logged in users. So separating the users via city on the index page then becomes a bit difficult.... The way I see it, there's 2 options:
1. show people organized by city for logged out users anyway (I don't really like this option thou, it feels a bit dishonest)
2. show just a standard list of ppl for logged out users & a list organized by city just for logged in users
3. something else

thoughts?

PragTob:
phew tough not to crack indeed. We could also do:

[...]

PR #507, opened 2017-02-07
add alphabetical jump links to people/coach index pages

berlintam:
This is semi-resolving issue #370 
I added alphabetical jump links for the people and coaches index pages, exciting stuff! I also refactored a bit as there was a lot of repetition between the people#index and the coach#index. I pulled a bunch of it out into partials. 
I'm only working in the non-redesign area thou, as I don't want to touch the redesign stuff until it's up and running & for realz. But since this is mostly in partials, switching it over to the redesign when ready shouldn't be much of a problem. 

I have no specs, as it's mostly display stuff & I don't really want to write a feature spec for this. But maybe a little bit of a controller spec? Thoughts?

Anyway, here's a screenshot:
screenshot from 2017-02-07 16-35-29

As always, review & comments always welcome!

### DIFF
--- a/app/assets/stylesheets/custom/_groups_people.scss
+++ b/app/assets/stylesheets/custom/_groups_people.scss
@@ -71,3 +71,16 @@
   border: solid 1px #B12200;
   border-radius: 5px;
 }
+
+.alphabetical-list {
+  a {
+    font-size: 1.3em;
+    padding: 0.2em;
+  }
+  margin: 1em 0 1em 0;
+}
+
+.letter-header {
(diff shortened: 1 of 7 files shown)


lislis:
This is awesome! Write a test if you want but I think this is great already! Especially how you generate the alphabetical list :D

berlintam:
woot woot! thanks @lislis. ya, I was pretty stoked about it as well :smile:

berlintam:
giphy

til:
That looks great, someone should deploy!

PR #508, opened 2017-02-14
Searching statuses for people and groups, closes #287

### LINKED ISSUE #287: [Feature] "Currently searching" status
Hello again everyone,

we realized that a lot of people have registered for the Rails Girls App but for a long while nothing happens after the registration. I assume that the first step to contact others is always the hardest for newbies, if they don't know the other people and their demands (searching/not searching).

An option to make it visible that someone is looking for something might help?

If yes, I would like to suggest following:
- provide an option to change a profile status into "currently searching" 
- after "searching"status is activated, provide a checkbox to select: a) project group b) coach c) co-students

Maybe this could be an additional feature to your ticket #115 to filter people by the status "searching"?

Happy to receive your feedback. :)

Best,
Thuy

lislis:
WIP

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -38,7 +38,7 @@ class GroupsController < ApplicationController
   def update
     if @group.update_attributes group_params
       redirect_to group_path(@group), notice: 'Group was successfully updated. ' \
-      'All efforts, nomatter how small, deserve cake.'
+                                              'All efforts, nomatter how small, deserve cake.'
     else
       render action: 'edit'
     end
@@ -58,7 +58,7 @@ class GroupsController < ApplicationController
   def destroy
     @group.destroy
     redirect_to groups_path, notice: 'Group was successfully deleted. ' \
-    'May we suggest drowning your sorrows in some cake?'
+                                     'May we suggest drowning your sorrows in some cake?'
   end
 
   private
@@ -68,9 +68,29 @@ class GroupsController < ApplicationController
   end
 
   def group_params
-    params.require(:group).permit(:id, :name, :address, :time,
-    :picture, :twitter, :contact, :activities, :email, :level,
-    :founded_on, :join_as_coach, :full, :city, :country, :street, :zip,
-    :learning_resources, :inactive, :allow_male_students)
-  end 
+    params.require(:group).permit(
+      :id,
+      :name,
+      :address,
+      :time,
+      :picture,
+      :twitter,
+      :contact,
+      :activities,
+      :email,
+      :level,
+      :founded_on,
+      :join_as_coach,
+      :full,
+      :city,
+      :country,
+      :street,
+      :zip,
+      :learning_resources,
+      :inactive,
+      :allow_male_students,
+      :searching_coaches,
+      :searching_students,
+      :searching_location)
+  end
 end
--- a/app/helpers/groups_helper.rb
+++ b/app/helpers/groups_helper.rb
@@ -17,4 +17,18 @@ module GroupsHelper
     [group.street, group.zip, group.city, group.country_name].select {|x| x.present? }.join(', ')
   end
 
+  def group_searching_fors(group)
+    fors = []
+    if group.searching_coaches?
+      fors.push('coaches')
+    end
+    if group.searching_students?
+      fors.push('students')
+    end
+    if group.searching_coaches?
+      fors.push('place to meet')
+    end
+    fors
+  end
+
 end
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -79,4 +79,8 @@ class Group < ActiveRecord::Base
   def capitalize_fields
     self.city = city.titleize
   end
+
+  def searching?
+    searching_coaches || searching_students || searching_location
+  end
 end
--- a/app/views/groups/show.html.haml
+++ b/app/views/groups/show.html.haml
@@ -124,3 +124,11 @@
             = link_to coach.first_name, person_path(coach), class: 'list--people-name'
             - if admin?
               = render partial: 'remove_from_group', locals: { object: coach, group: @group }
+
+      - if group_searching_fors(@group).any?
+        %h3 This group is
+        %p
+          - group_searching_fors(@group).each do |searching_for|
+            %span.tag
+              searching for
+              = searching_for
--- a/app/views/people/show.html.haml
+++ b/app/views/people/show.html.haml
@@ -32,7 +32,9 @@
       %p
         Your account is not public
         %small (only registered users can see your profile)
-
+    - if @person.searching_group?
+      %p
+        %span.tag Searching for a group!
 
   .profile-main
     - if logged_in?(@person)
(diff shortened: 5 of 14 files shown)


lislis:
Rebased to current design

lislis:
I made the scope of this PR smaller. There will 'just' be tags that can be added, no searching or filtering by them yet.

I'll open another issue for this.

Here are some screenshots though

screenshot at 2017-04-25 20-20-54
screenshot at 2017-04-25 20-21-21

@til can you review?

app/controllers/groups_controller.rb
```diff
@@ -71,7 +71,8 @@ def group_params
     params.require(:group).permit(:id, :name, :address, :time,
     :picture, :twitter, :contact, :activities, :email, :level,
     :founded_on, :join_as_coach, :full, :city, :country, :street, :zip,
-    :learning_resources, :inactive, :allow_male_students)
+    :learning_resources, :inactive, :allow_male_students, :searching_coaches,
+    :searching_students, :searching_location)
```
til: Every line below `params.require` should be indented one level more, like this:

```ruby
params.require(:group).permit(:id, :name, :address, :time,
  :picture, :twitter, :contact, :activities, :email, :level,
  :founded_on, :join_as_coach, :full, :city, :country, :street, :zip)
```

I know it was wrong before your changes, but I thought I'd mention it anyway. Maybe a good opportunity to fix it. I'd also consider putting each attribute on one line - it will get a bit long then, but easier to read.

app/models/group.rb
```diff
@@ -73,4 +73,8 @@ def active?
   def location
     [address, street, zip, city, country].join(' ')
   end
+
+  def is_searching?
+    searching_coaches || searching_students || searching_location
+  end
```
til: The ruby style guide says, call it `searching?` instead: https://github.com/bbatsov/ruby-style-guide#bool-methods-prefix

app/views/groups/show.html.haml
```diff
@@ -124,3 +124,13 @@
             = link_to coach.first_name, person_path(coach), class: 'list--people-name'
             - if admin?
               = render partial: 'remove_from_group', locals: { object: coach, group: @group }
+
+      -if @group.is_searching?
```
til: space missing after the `-`

app/views/groups/show.html.haml
```diff
@@ -124,3 +124,13 @@
...
+        %h3 This group is
+        %p
+          - if @group.searching_coaches
+            %span.tag searching for coaches
+          - if @group.searching_students
+            %span.tag searching for students
+          - if @group.searching_location
+            %span.tag searching for a place to meet
```
til: Whenever possible, the variant of the method name with the question mark should be used, so e.g. `searching_coaches?`

til: Theoretically this logic would be a good candidate to move into a helper or something similar, so that some duplication and logic can be removed from the view. The call in the view could then look e.g. like this:

```haml
- if @searching_fors.any?
  %h3 This group is
  - @searching_fors.each do |searching_for|
    %p
      %span.tag= searching_for
```

just an idea though ...

app/views/people/show.html.haml
```diff
@@ -32,7 +32,9 @@
       %p
         Your account is not public
         %small (only registered users can see your profile)
-
+    -if @person.searching_group?
```
til: space

app/views/shared/_grouped_people.haml
```diff
@@ -9,6 +9,8 @@
         .list-group-item-info
           %h3
             = profile_link(person)
+          - if person.searching_group
```
til: question mark

db/migrate/20170214202206_searching_status.rb
```diff
@@ -0,0 +1,8 @@
+class SearchingStatus < ActiveRecord::Migration
+  def change
+    add_column :people, :searching_group, :boolean, default: false, null: false
+    add_column :groups, :searching_coaches, :boolean, default: false, null: false
+    add_column :groups, :searching_students, :boolean, default: false, null: false
+    add_column :groups, :searching_location, :boolean, default: false, null: false
+  end
+end
```
til: Not allowing null, good!

spec/features/searching_status_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      displays_searching_for_location_status
+    end
+  end
+
+end
+
+def does_not_display_searching_for_group
+  visit person_path(person)
+  expect(page).to_not have_content('searching group')
+end
```
til: Put those helper methods _inside_ the outermost spec level, ie. the block of `feature`. Otherwise I think they become global methods in all specs, this is not what we want.

spec/features/searching_status_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      go_to_group_settings
+      check_searching_for_students_and_save
+      displays_searching_for_students_status
+
+      does_not_display_seaching_for_location
+      go_to_group_settings
+      check_searching_for_location_and_save
+      displays_searching_for_location_status
+    end
+  end
```
til: Cool, very detailed spec! There's a typo repeated, "seaching" instead of "searching".

In a feature spec, the convention is to use `feature` + `scenario` instead of `describe` + `it`. I think using `context` is fine, although I have never needed it in a feature spec yet.

lislis:
@til  thanks for the review! I think I addressed all the things. But not 100% sure if the helper is what you meant?

app/controllers/groups_controller.rb
```diff
@@ -68,11 +68,30 @@ def set_group
...
+      :city,
+      :country,
+      :street,
+      :zip,
+      :learning_resources,
+      :inactive,
+      :allow_male_students,
+      :searching_coaches,
+      :searching_students,
+      :searching_location)
```
til: Good. Maybe do the same for PeopleController as well?

app/helpers/groups_helper.rb
```diff
@@ -17,4 +17,18 @@ def group_location(group)
...
+    end
+    if group.searching_students?
+      fors.push('students')
+    end
+    if group.searching_coaches?
+      fors.push('place to meet')
+    end
+    fors
+  end
+
```
til: Now that I see it I recall another convention, which is that helper methods should start with the area that they are in, to make clear where they belong. That's because in Rails, helper methods all are put into the same quasi global helper namespace. In this case, this would probably mean to call it `group_searching_fors` ... it sounds a bit weird, so if you or someone has a better idea, feel free to change it, but until then, it's probably better to at least follow the convention.

til:
> But not 100% sure if the helper is what you meant?

Besides the inline comments: yes, that's what I meant. This way the
logic is moved out of the view, which is how it should be - only the
most simple loops and if's should be in views.

app/controllers/groups_controller.rb
lislis: sure

app/helpers/groups_helper.rb
lislis: +1

til:
:+1:

lislis:
grinnnnnn

PR #512, opened 2017-02-21
paginate person & coach index pages

berlintam:
Final step in https://github.com/rubycorns/rorganize.it/issues/370

I added pagination to the people#index & coaches#index, which is also represented in the alphabetical jump links & also works based on the filtering selected. See screenshots for examples. 

One could argue that there's some repetition in the controllers... any suggestions for how to improve this (or anything else of course) is very welcome. I also tried to be super explicit about naming my instance variables in the controllers, so I hope they're not confusing. If they are, please let me know. 

I also thought that maybe my code to add `'...'` to the beginning and end of the alphabetical list [here](https://github.com/rubycorns/rorganize.it/blob/paginate-user-page/app/views/shared/_alphabetical_list.haml#L2) could be moved elsewhere... but I'm not sure where. In other projects I've worked on we've used decorators, and I think that would be a good place for this but we don't use decorators. So... hmmmmmm.....

#### page 1
screenshot from 2017-02-21 14-17-29

#### page 2
screenshot from 2017-02-21 14-17-55

#### using filter (Berlin selected)
screenshot from 2017-02-21 14-18-13

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -343,7 +343,7 @@ GEM
       addressable (>= 2.3.6)
       crack (>= 0.3.2)
       hashdiff
-    will_paginate (3.0.5)
+    will_paginate (3.1.5)
     win32console (1.3.2-x86-mingw32)
     xpath (2.0.0)
       nokogiri (~> 1.3)
@@ -404,4 +404,4 @@ DEPENDENCIES
   will_paginate
 
 BUNDLED WITH
-   1.11.2
+   1.14.3
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,5 @@
 class ApplicationController < ActionController::Base
+  require 'will_paginate/array'
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
--- a/app/controllers/coaches_controller.rb
+++ b/app/controllers/coaches_controller.rb
@@ -1,16 +1,20 @@
 class CoachesController < ApplicationController
   def index
-    ordered_coaches = Person.workshop_coach.order_by_name
-    coaches_by_region = ordered_coaches
-                          .filtered_by_region(params)
-                          .filtered_by_visibility(signed_in?)
+    coaches = Person.workshop_coach.order_by_name
+                .filtered_by_region(params)
+                .filtered_by_visibility(signed_in?)
 
     @willing_to_travel = params[:willing_to_travel] == '1'
-    @coaches = @willing_to_travel ? coaches_by_region.willing_to_travel : coaches_by_region
+    @coaches = @willing_to_travel ? coaches.willing_to_travel : coaches
     @cities  = Person.workshop_coach.visible_locations_for(:cities, signed_in?)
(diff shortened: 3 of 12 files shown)


berlintam:
OH, and edge case question: we validate for first names. But what if somehow a person doesn't have a first name? Should I cover this case or is it very unlikely?

til:
Hm, the alphabetical grouping and the paging look like two orthogonal criteria that don't really make sense to mix together, in my very humble opinion. It seems weird that only a part of the alphabet is visible in the beginning, and one needs to page to get to the other part. The list of possible letters (if restricting it in a culturally-insensitive way to the 26 ascii chars) is always fixed and would always fit the width of a (desktop) screen, so it would be a good candidate for a grouping criteria I think - meaning that once we have thousands of users, we could start showing only those who fall into the letter "a", and make the jump links actual page links to switch to the subset of another letter.

Up to how many users can we show at once at one page - any estimates? I would say something like 300 would still be ok, but that's just a totally unempirical gut feeling.

berlintam:
Thanks @til!
> The list of possible letters (if restricting it in a culturally-insensitive way to the 26 ascii chars) 

So, currently it's not limited to 26 characters. We display whatever the first letter of the name is in the alphabetical list. Maybe alphabetical is the wrong word for this list as it's more a 'character' list. If someone's first name starts with `$`, then it'll be shown there as well. I have a spec for that here https://github.com/rubycorns/rorganize.it/blob/835fcd0c4f1dce11744282e1f128dbca6676bf2b/spec/models/person_spec.rb#L276

> Up to how many users can we show at once at one page - any estimates? I would say something like 300 would still be ok, but that's just a totally unempirical gut feeling.

I limited it to 30 ppl, but we can totally change it. Do you mean how many users can be show at once without noticeable lag? That I'm not sure of..... How would one test that?

So my suggestions for fixing this are: 
* showing the character list on every page, and if the first character of your name falls on another page, clicking on that character jumps you to the right page. 
* clicking on a character brings you to an entirely new page. One character per page. 
* showing the character list on every page and implement autoscrolling instead of pagination! (is that what it's called? Whatever twitter/fb does...)

Thoughts?

lislis:
I would like to have all character links on every page for overview reasons (I also don't expect too many people whose names start with $ or #).
 
For listing: I think we can use some CSS trickery here ❇️ 

There is a pseudo selector `:target` that only applies styles if the element is 'targeted' (http://someurl.com#target).
The main idea is that we hide all sections that are currently not the target. If you click on a letter, the section with the corresponding `id` attribute becomes the target and the `:target` styles then display the section.

Wdyt? Does that make sense?

til:
I must admit I'm a little bit confused by all of this, but after having @lislis explain the `:target` approach to me so that I understand it now, I think it would be the best and easiest way forward.

The html page stays what it is right now - a full list of all people. There are still not too many users, so this should work. The page loads a bit slow right now, ~2s, but maybe that can be improved separately.

@berlintam would that be ok for you? Is there anything in this PR that would still make sense to port over?

berlintam:
Just talked w/ Lisa about this (yes, I am at the corns!) and I'm going to figure this out with target instead, so I'm closing this PR.

PR #514, opened 2017-02-25
Updating Master with Redesign Branch

sareg0:
This is to ensure Master has the latest updates, so we can all help get the redesign out, but pulling master


til:
Looks good, tests passing? 3 ... 2 ...

Issue #523, opened 2017-02-25
Group single view: Show resources/topics/members as tabbed content


katrin-k:
Resources, topics and members are currently shown below each other. The plan is to have them shown as tabs. 

In groups#show and main.js, there is already Markup and the JS, which is currently outcommented.

PR #528, opened 2017-02-26
Group profile: Tabbed content for resources, topics and members

katrin-k:
Issue #523  

Problem: The topics tab is not shown when there are no topics, but the user is member of a group and is allowed to start topics.

### DIFF
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -1,35 +1,20 @@
-function makeGif() {
-	if ($('#gifModal').length) {
-		randomGif.init();
-		$('.make-tooltip').tooltip();
-	}
-}
-
-$(document).on('page:load', makeGif);
-
 $(document).ready(function() {
 
-	function removeAlert() {
-		$('.alert').addClass('fade-out');
-	}
-	window.setTimeout(removeAlert, 8000);
+  function removeAlert() {
+    $('.alert').addClass('fade-out');
+  }
+  window.setTimeout(removeAlert, 8000);
+  
+  // initiates the randomGif maker that we use
+  // in the admin panel
+  if ($('#gifModal').length) {
+    randomGif.init();
+  }
 
-	makeGif();
-
-	// ==> Tabs (see groups#show)
-	// 1. Hide all tabs except first one when page loads
-	// $(".tab").addClass('hidden');
-	// $(".tab:first").show();
-	//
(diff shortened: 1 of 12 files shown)


app/views/groups/show.html.haml
```diff
@@ -81,46 +81,43 @@
...
-    -#     %a{:title => 'Members', :href => '#members'} Members
+    %ul.group-tabs
+      - if @group.learning_resources?
+        %li.group-tab
+          %a{:title => 'Resources', :href => '#resources'} Learning Resources
+      - if @group.topics.any?
+        %li.group-tab
+          %a{:title => 'Topics', :href => '#topics'} Topics
+      %li.group-tab
+        %a{:title => 'Members', :href => '#members'} Members
```
til: I know it's been like this before, but it would be nice to change it to use the modern hash syntax here (`title: 'Resources', ...`).

lislis:
@katrin-k I think I don't understand the problem?

I think I remember that we said that topics should be readable by everyone, but comments only by member of the group? 
So everyone should see the topics tab. 
Is that what you mean?

app/assets/javascripts/main.js
```diff
@@ -1,35 +1,34 @@
...
+
+function addTabActive() {
+  $('.group-tab a:first').addClass('active')
+  $('.group-tab a').click(function(e){
+    $('.group-tab a').removeClass('active');
+    $(this).addClass('active');
+    $('.tab.active').removeClass('active');
+    $($(this).attr('href')).addClass('active');
+  });
 }
```
til: It would be nice if you found a way to make these two functions not defined in the global scope.

sareg0:
LGTM

sareg0:
If a test fails everytime the build is run, can we still consider it a randomly failing test?

til:
I guess no :(

Can you reproduce it locally (when you have your private computer
available), and try to find out what is going on?

Besides the reason for the spec failure, what I also don't understand is
why it only happens in the push check, but not the pr check, which is
green. In my understanding the former is the code in the branch, the
latter is the code as it would look if it was merged into master. But
since we rebased on master, they should be the same, no? Or did
something else change in master in the meantime?

sareg0:
Here is a fancy screenshot of what this looks like:

screen shot 2017-07-04 at 21 35 24

Issue #536, opened 2017-04-15
Filter option ordering


dondonz:
Howdy,

At the moment, filter options for city and country are being displayed in the order they appear in the DB. Shall we specify an order, to make it easier to see if there is a group in a particular city or country?

For cities: how about we order these alphabetically?

For countries: shall we order this alphabetically, or by some other rule?

til:
Good point. I think any order would be better than the current one -
alphabetically also for countries is totally ok.

ZackTaylor:
@dondonz I'd like to take a shot at this. What controller is this in specifically? And should I order them alphabetically?

berlintam:
looks like this is done!

PR #538, opened 2017-04-20
Order filter options alphabetically (City and Country)

dondonz:
Hello, this is for #536. I've changed it so the filter options are displayed alphabetically for city and country.

I'm open to suggestions.

Regarding testing, is this something that would be worth writing a test for? Keen to hear your opinions on this.


lislis:
Awesome sauce, thanks @dondonz !
uuuuhh wheeee

Issue #535, opened 2017-04-15
City with more than one name


dondonz:
Hello,

In the Groups page, the City filter has two Berlins - one is lowercase ('berlin'), the other is capitalised ('Berlin'). I'd like to change it so that there can only be one name for every city.

I'm guessing there is a pre-save or pre-validation method for this. I'm happy to look into it. We will need to also manually change the 'berlin' groups to 'Berlin' from the console once this is complete.

til:
Cool, thanks for reporting and for looking into it! There really should
only be one Berlin.

lislis:
After talking to @til there are two things we need:

1) clean up the production database to only have each city once
2) write some code that prevents duplicate city names in the future

@dondonz would you like to tackle 2)?

dondonz:
Yes I'd love to!

lislis:
closed by https://github.com/rubycorns/rorganize.it/pull/542

PR #542, opened 2017-04-26
Capitalizing City Name Before Save - in progress

dondonz:
This Addresses Issue #535.

In The Group Model, I've Added A `before_save` Method To Capitalize A City's Name Before Saving To The Database.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -72,6 +72,5 @@ class GroupsController < ApplicationController
     :picture, :twitter, :contact, :activities, :email, :level,
     :founded_on, :join_as_coach, :full, :city, :country, :street, :zip,
     :learning_resources, :inactive, :allow_male_students)
-  end
-
+  end 
 end
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -49,6 +49,8 @@ class Group < ActiveRecord::Base
   geocoded_by :location
   after_validation :geocode
 
+  before_save :capitalize_fields
+
   extend FriendlyId
   friendly_id :name, use: :slugged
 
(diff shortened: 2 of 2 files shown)


dondonz:
Next up - update the method for cities with 2 word names (eg New York)

til:
Cool, thanks for starting with this!

One minor style remark - the extra empty lines before the end of a class are not necessary. The ruby style guide says they should not be added: https://github.com/bbatsov/ruby-style-guide#empty-lines-around-bodies - I'm sure though we are totally not consistent with this, and have a mixture of all styles in our current codebase.

dondonz:
@til Thanks! I've removed the line breaks before the end.

I've changed .capitalize to .titleize. If a user types 'berlin' or 'new york' it will now be converted to 'Berlin' or 'New York'.

I aimed to solve the case where a user accidentally types in the lower case form of the name. Of course, there are infinite possibilities for misspelling:

- If a user types 'new yoRk', the name will become 'New Yo Rk'.
- If a user types 'Berlin2', the name will still be 'Berlin2'

And... another edge case:
- If a user lives in a town with a hyphenated name, 'Place-Name' will become 'Place Name'

What do you think - shall we stop at .titleize?

til:
`titleize` - good find. There seems to be a method for everything in the
Rails code.

I think this solution is great. Looking at the list of cities we have
currently, there is nothing exotic in it. One day we might get a problem
when someone enters a city with a dash (do those even exist?), but we
might look into that once it happens, if ever.

Do you want to squash and rebase onto master?

dondonz:
Fantastic! Thanks @til!

lislis:
Awesome!
I cleaned up the data on the production database. I'll merge this and deploy!

lislis:
AAAAAnd deployed!
wheee

til:
Great!

PR #544, opened 2017-05-09
Add group admin, close #487

### LINKED ISSUE #487: Rethink group editing permissions
We got a feedback email concerning who has permissions to edit a group.
Here is a short summary (abbreviated original converstion):

```
A: it seems that currently everyone is allowed to edit groups. Is that intended?
B: Currently every member of a group can edit that group.
A: I would prefer if only the owners of the group could do that
B: How would you define the owners?
A: The creator of the group should be the owner. But it then would be probably necessary to give ownership also to someone else.
```

Other opinions?

lislis:
Add the concept of group admins.
Only a group admin can edit the group info.
A group admin can assign more group admins.
If a person creates a new group they automatically become that group's admin


lislis:
What I'm not sure about, how will we treat existing memberships? I chose default false but I guess to not break anything all existing ones need to be set to admin: true?

lislis:
Also I'm not sure how to actually get the admin flag off a membership? `person.memberships` yields all of them but I want the one of a particular group

app/controllers/groups_controller.rb
```diff
@@ -24,9 +24,11 @@ def create
     @group = Group.new(group_params)
     if @group.save
       if @group.join_as_coach == '1'
-        current_person.join!(@group, 'CoachMembership')
+        current_person.join!(@group, 'CoachMembership', true)
+      else
+        current_person.join!(@group, 'StudentMembership', true)
       end
-      redirect_to group_path(@group), notice: 'Group was successfully created. This calls for cake!'
+      redirect_to group_path(@group), notice: 'Group was successfully created. You are the admin now. This calls for cake!'
```
til: 🎵 Oooh ho ho you are the admin now

(thanks, I'll show myself out ...)

app/models/person.rb
```diff
@@ -105,7 +105,7 @@ def to_s
     full_name
   end
 
-  def join!(group, type = 'StudentMembership')
+  def join!(group, type = 'StudentMembership', admin = false)
```
til: Let me suggest a readability improvement: change all optional arguments to be keyword arguments. This way, when calling it it would read e.g. like this:

`join!(@group, admin: true)`

this way the reader does not have to think what the boolean argument might mean. Also it gives the possibility, like in this example, to leave the other optional argument `type` at its default without specifying it.

You could also just change the `admin` argument to a keyword argument, but when you're at it why not also change `type`, and improve it compared to how it was before.

db/migrate/20170509182902_add_admin_to_membership.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdminToMembership < ActiveRecord::Migration
+  def change
+    add_column :memberships, :admin, :boolean, default: false, null: false
```
til: 👍 for default and not null

db/schema.rb
```diff
@@ -113,14 +114,14 @@
...
-    t.string   "picture"
+    t.string   "picture",      limit: 255
     t.datetime "created_at"
     t.datetime "updated_at"
     t.integer  "person_id"
-    t.boolean  "draft",        default: false, null: false
-    t.string   "slug"
+    t.boolean  "draft",                    default: false, null: false
+    t.string   "slug",         limit: 255
     t.datetime "published_at"
```
til: Many unrelated changes in everybody's darling file, the schema.rb. Why does it add "limit: 255" to all strings now? Maybe a new rails version, or a different database? If it is from a new rails version, then it should be made permanent, but it'd be much better as a separate commit - maybe you can employ some hardcore git rebasing to get those changes out?

lislis: so If I check out the schema from master and run `rake db:mirgate` again, it adds all the limits.
I use the rails version that is specified in the Gemfile.lock on master. Not sure if it's a db thing?

lislis:
Ok, so I added a field on the person model that checks if the person is admin member of the group.
Now I'm hiding the group edit button on group#show for everyone who is not a group admin or a big overall admin.
What I'm missing is to restrict access to actually editing the group

db/schema.rb
til: Same for me when I check out master and run `rake db:migrate`. Then it is something that we should have changed already, I suppose. I think the best way would be to add these limit changes in a separate commit to master, but I won't complain if it gets added together with the commits from this PR.

lislis: I have the changes to schema in a commit. maybe we can cherry-pick that to master?

lislis:
I solved the edit problem by introducing another `before_action` that checks for general admin or group admin powers.
Todo:
- [ ] fix tests
- [ ] add group dashboard where group admins can add more group admins
- [ ] think about how to migrate existing data (eg make all current members of groups admin members)

lislis:
Added some UI for the group admin to interact and user manage
selection_010

app/controllers/memberships_controller.rb
```diff
@@ -14,6 +14,12 @@ def create
     end
   end
 
+  def update
+    membership = Membership.find(params[:id])
+    membership.update_attribute(:admin, !membership.admin)
+    redirect_to group_path(membership.group), notice: "Successfully changed membership status. Just like icing on the cake."
+  end
+
```
til: This definitely needs a permissions check, otherwise everyone can make everyone admin. Probably `ensure_group_admin_powers` can be used for that, after some adaptation. What would be great: a spec that ensures that the security check is working - I think a controller spec would be appropriate for this.

til: Nice notice text 😆

spec/features/group_member_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    check_group_admin
+    can_edit_group
+  end
+
+  scenario 'Join existing group and not be admin of it' do
+    sign_in_user
+    join_group
+    check_not_group_admin
+    can_not_edit_group
+  end
```
til: 👍 this reads very well imho.

A minor remark, maybe use the same naming scheme like in `can_not_edit_group` also for `check_not_group_admin` and call it instead e.g. `can_manage_members`.

A pair of methods like `can_edit_group` and `can_not_edit_group` must always stay in sync, otherwise there is the danger of missed failures, e.g. when the link text is changed only in one but not the other. I suggest to move them around so that each pair is next to each other. Another safeguard is to pull the link text into its own mini-method, so that `can_edit_group` becomes

```ruby
def can_edit_group
  expect(page).to have_content(edit_group_link_text)
end
...
def edit_group_link_text
  "Edit group"
end
```

Even more sophisticated would be a custom matcher that allows to say e.g. `expect(page).to allow_edit_group` which then can be simple negated to `expect(page).to_not allow_edit_group` but I have to admit I'm not using that as much, even though I probably want to.

app/controllers/memberships_controller.rb
til: Bump

lislis:
the only two tests failing seem to be related to the non-existence of `current_person` in the groups controller. This might be a 🐰 🕳 I don't want to fall into tonight

lislis:
Memo to myself: check permissions on the memberships destroy and update methods

lislis:
I tried checking for permissions but that seems to break more tests 🤦‍

lislis:
I removed what I tried to do on the memberships controller but was able to fix the tests on the groups controller.

lislis:
@til would you be ok with shipping this and then looking at the membership controller permission in a different PR?

db/migrate/20170509182902_add_admin_to_membership.rb
```diff
@@ -0,0 +1,10 @@
+class AddAdminToMembership < ActiveRecord::Migration
+  def up
+    add_column :memberships, :admin, :boolean, default: false, null: false
+    Membership.all.each { |m| m.admin = true }
```
til: This won't update it in the database, because it doesn't save. Either call `save!`, or use http://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-update_all

db/seeds.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
til: Hello emacs ;-)

spec/features/group_topics_spec.rb
```diff
@@ -28,7 +28,7 @@
   def sign_in_user
     visit new_person_session_path
     sign_in person
-    person.join!(group, 'StudentMembership')
+    person.join!(group, {})
```
til: The `join!` method could be written in a way that the default empty hash is not needed, so that in this case it just be called like this: `person.join!(group)`. The method definition could take named keyword args (yay modern Ruby!), and also define the defaults with it.

spec/features/searching_status_spec.rb
```diff
@@ -21,7 +21,7 @@
     let(:person) { create(:person) }
     let(:group) { create(:group) }
 
-    before { person.join!(group, 'StudentMembership') }
+    before { person.join!(group, {admin: true}) }
```
til: Once join! uses keyword arguments, this could be written as `person.join!(group, admin: true)`. Probably that already works now.

app/controllers/groups_controller.rb
```diff
@@ -1,10 +1,11 @@
 class GroupsController < ApplicationController
   respond_to :html
 
-  before_action :set_group, only: [:show, :edit, :update, :destroy]
+  before_action :set_group, only: [:show, :edit, :update, :destroy, :manage_members]
   before_action :authenticate_person!, except: [:index, :show]
   before_action :ensure_admin_powers, only: [:destroy]
   before_action :ensure_member_powers, except: [:index, :show, :new, :create, :destroy]
+  before_action :ensure_group_admin_powers, only: [:edit, :destroy, :manage_members]
```
til: There is some overlap between the permissions, e.g. do I need to be both admin and group admin for destroy? What about update - shouldn't it be restricted too?

This is quite important, not that by accident something is allowed that shouldn't ...

lislis: technically you would need to be either an admin or a group admin to :destroy how would you write it?

db/seeds.rb
lislis: lol

til:
So, regarding whether to make everyone an admin or only coaches: there are 4 groups without coach on production, of which 2 are active:

```
irb(main):018:0> Group.all.select {|g| g.coaches.empty? }.map(&:name)
=> ["Rails Girls G��", "Ruby Barna", "wednesday Ruby", "RubyDoobyDoo"]
irb(main):019:0> Group.all.select {|g| g.coaches.empty? }.map(&:active?)
=> [true, true, false, false]
```

An alternative would be to give only the person with the first membership (by created_at) the admin permission. A little complicated for the migration, but should be possible. I checked, all memberships seem to have reasonable created_at entries.

app/controllers/groups_controller.rb
til: Put the logic into a private method, and use that as before_action, e.g. `before_action :ensure_can_destroy, only: [:destroy]`

lislis:
it's happening

berlintam:
Such excite!! Thanks for doing this Lisa!

PR #546, opened 2017-05-16
Change looks of blog#index

lislis:
I put some styles on the index of the blog to hopefully facilitate the overview

selection_005

The rubycorn gif is for posts with no thumbnail, btw


app/views/posts/show.html.haml
```diff
@@ -6,7 +6,24 @@
...
     = link_to 'Edit post', edit_post_path(@post), class: 'btn btn-teal'
     = link_to 'Delete post', @post, method: :delete, data: {confirm: 'Are you sure?'}, class: 'btn btn-pink'
-= render partial: 'post', object: @post
+
+%article.post.post-single
+  %header
+    - if current_person.admin? && @post.draft?
+      %p.post-draft
+        It's a draft!
+    %h2.post-title= link_to "#{@post.try(:title)}", post_path(@post)
```
til: Is the string interpolation necessary here?

And also is the `try` necessary - when would `@post.title` not work - only when `@post` is nil, or is there another reason why you found it necessary? Ideally, but I am not sure about that, it should be able to just call `link_to @post` and Rails does the right thing. But maybe that only works in a specific situation, or another version of Rails. Maybe it's necessary to define `to_s` in `Post` for that.

app/views/posts/show.html.haml
```diff
@@ -6,7 +6,24 @@
...
+    %p.post-meta
+      = @post.person.first_name if @post.person
+      |
+      \#{show_date(@post)}
```
til: Would `= show_date(@post)` also work? If yes I think that would be a bit simpler.

app/views/posts/show.html.haml
lislis: was wondering why the interpolation was there in the first place. It seems to work just fine without them so, I'll delete them

sareg0:
LGreatTM!

Issue #531, opened 2017-04-06
Signup from GitHub assumes last word of fullname is surname


porras:
And the rest to be the first name, which only works for some languages/cultures (German and English, for example) but doesn't for many (Spanish or Portuguese speaking, for example):

<img width="178" alt="screen shot 2017-04-06 at 15 18 55" src="https://cloud.githubusercontent.com/assets/3830/24756107/e9b1fa7a-1adc-11e7-83ea-ed908dd72259.png">

(My first name is Sergio and the rest is my surname).

The code that does that assumption seems to be [here](https://github.com/rubycorns/rorganize.it/blob/master/app/models/person.rb#L92-L96).

I don't know what would be the solution though, it's a tricky problem to solve for all 😕

dondonz:
This is tricky...!

If I can add another complexity, in many East Asian languages/cultures (eg Chinese) and a few others (eg Hungarian), a person would state their full name as family name first, followed by the given name(s).

lislis:
Mhhh, tricky indeed. Maybe in the code that @porras pointed out, we should change the logic to use `names.first` as first name and join the rest as last name (the other way around from what it is now.).

That would annoy people with two first names that are not combined with a `-` but I think is overall more sensible.

We could also write a note that we assume Central European/ English language style names and are sorry for when it doesn't work out?

vsmart:
Fixed in https://github.com/rubycorns/rorganize.it/pull/550

PR #550, opened 2017-05-30
Adresses #531 re importing profile names from Github.

vsmart:
The aim is to improve name import from Github
This solution, as suggested by @lislis, is not perfect but I think slightly better than previous; assume the first name is the first name and all others are last name.


lislis:
yisss

PR #551, opened 2017-05-31
Update to Ruby 2.4; Rails 4.2.8 - fixes #549

### LINKED ISSUE #549: Deployment not working
Pushing to production master does not seem to kick off deployment at the moment.

vsmart:
This should fix https://github.com/rubycorns/rorganize.it/issues/549 !


til:
Thanks for fixing this!

:+1: - can you merge and deploy?

I think sometime later we could get rid of the exact version declaration
in the Gemfile again (ie, not setting it to exactly "4.2.8"). In my
opinion the fewer exact version specifications there are in the Gemfile
the better, they are all locked in the Gemfile.lock anyway. Many
projects seem to do it differently, but I have never understood why.

Also another thing for later (feel free to do that if you want), we
should also specify the ruby version in the Gemfile:
https://bundler.io/v1.15/gemfile_ruby.html

vsmart:
@til I can merge, someone else needs to deploy though :) (I don't have uberspace access)

lislis:
sweet, I'll deploy right now
woooorkin hard

PR #555, opened 2017-06-06
Add autostart script to git

lislis:


til:
👍 

if someone else wants to deploy: please do, but make sure that the currently not checked in `config/autostart` on production is not in the way, maybe by removing it manually before doing the deploy.

lislis:
I deleted the `config/autostart` form uberspace and push current master.
After a few moments of downtime we app restarted. 
Seems to work. YAY
working

vsmart:
cat yay

til:
Yaaaaay, well done! 🚀

PR #560, opened 2017-08-14
Updated CSS to fix regression in top bar

m-gb:
I removed the display:block in the profile-header CSS class.
The avatar and user name are now aligned.


app/assets/stylesheets/custom/_layout.scss
```diff
@@ -150,10 +150,7 @@ nav.navbar {
     align-self: center;
   }
   &.person-profile {
-    display: none;
-    @media (min-width: $mq-large) {
-      display: block;
```
lislis: The easiest would be to just change this line to `display: flex`

m-gb:
Thank you for responding, I didn't know that was the expected behavior.
I updated the pr, let me know if it's ok.

til:
Great, thanks! Yes, this way is the right one, the navbar looks good now.

I reverted the change to the `.ruby-version` file and squashed the commits into one, which is the reason why your commit is now in master: https://github.com/rubycorns/rorganize.it/commit/09dfa2434ccc00e6af3d1e58fa50c1371c868047 even though I'm just closing this PR now without merging.

Also it's deployed live: https://rorganize.it/

lislis:
yaaay

PR #561, opened 2017-08-15
Setup error reporting 363

sareg0:
This is a basic first implementation for the exception notifier. I'm not sure what else we would want to receive notifications on apart from exceptions. Contribution and review on this PR is welcome and encouraged 🙏 Many thanks in advance for your thoughful feedback.


lislis:
As reference and help, I used the same gem here https://github.com/rubymonsters/slam-alphas/pull/198/  @sareg0

til:
And another hint: the exception_notification now supports error grouping to limit the number of mails received in case of an error: https://github.com/smartinez87/exception_notification#error-grouping

It needs to be explicitly turned on, and definitely should be.

config/environments/development.rb
```diff
@@ -28,6 +28,12 @@
...
 
-  config.action_mailer.delivery_method = :test
-
+  config.action_mailer.delivery_method = :letter_opener
+
+  config.middleware.use ExceptionNotification::Rack,
+    email: {
+      email_prefix: "[ERROR] ",
+      sender_address: %{"Rubycorns" <roar@rorganize.it>},
+      exception_recipients: %w[regan.sara@gmail.com]
```
PragTob: maybe change me or remove me? :)

lislis:
party

PR #564, opened 2017-09-16
Move students list up to proper position under students list.

ZackTaylor:
Move students list to proper place in the groups#show view.


lislis:
thumbs up
Thank you!

PR #565, opened 2017-09-26
display alphabetical group when clicked (pagination)

berlintam:
retry for https://github.com/rubycorns/rorganize.it/pull/512

slight problem is, if we have an alphabetical list (from A-Z) and go to /people, we display the people whose names start with Z first, not the ppl whose names start with A. 

@sareg0 is gonna take another look at this! suggestions welcome on how to get the page to display at A group first :)

(ex: in the screenshot you can see that if you go to /people the last alphabetical group of people is shown first)

screenshot from 2017-09-26 21-57-23

### DIFF
--- a/app/views/people/index.html.haml
+++ b/app/views/people/index.html.haml
@@ -11,3 +11,10 @@
 
   %aside.flex-index-sidebar
     = render partial: 'shared/filter', locals: { path: people_path }
+
+
+%script
+  :plain
+  if (window.location.hash === '') {
+    window.location.hash = "#{@alphabetical_list.first}";
+  }
--- a/app/views/people/index.html.haml
+++ b/app/views/people/index.html.haml
@@ -13,8 +13,8 @@
     = render partial: 'shared/filter', locals: { path: people_path }
 
 
-%script
-  :plain
+:javascript
   if (window.location.hash === '') {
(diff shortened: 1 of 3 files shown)


app/views/people/index.html.haml
```diff
@@ -11,3 +11,10 @@
 
   %aside.flex-index-sidebar
     = render partial: 'shared/filter', locals: { path: people_path }
+
+
+%script
+  :plain
```
til: There is also a `:javascript` thing - filter I think it is called - in haml, which outputs a script tag automatically.

app/views/people/index.html.haml
```diff
@@ -11,3 +11,10 @@
...
+  if (window.location.hash === '') {
+    window.location.hash = "#{@alphabetical_list.first}";
+  }
```
til: Also, not only GitHub prefers a file to end in a newline

berlintam:
Thanks for taking this over @sareg0!! :green_heart:

sareg0:
thanks @til! and @berlintam you're welcome! Just filling in the lil' bug you were having to close this puppy up! 🐩

lislis:
puppies

til:
Time to mööörrrge this!!

berlintam:
excite!

PR #567, opened 2017-10-29
Fix signup form to autofocus first input field

cfleschhut:
### Small usability fix:
Signup form in `registrations#new` view should initially focus the first input field, not the third one

#### Before

bildschirmfoto 2017-10-29 um 14 41 17

#### After

bildschirmfoto 2017-10-29 um 15 49 20


lislis:
Hey there @cfleschhut!
Thank you very much!
thumbs up!
Sorry for the failing tests though, we're leaking some requests :/ Running it again made it pass

Issue #519, opened 2017-02-25
Responsive styles


katrin-k:
Check styles for phone, tablet and mega-screens.

sareg0:
We took a quick look, and spotted some good first candidates for fixing up

- navigation needs to have some sort of mobile friendly menu
- better stacking on most pages, so that content is visible and/or not squished

Some examples below

<img width="291" alt="screen shot 2017-03-21 at 19 49 24" src="https://cloud.githubusercontent.com/assets/1307818/24165022/a26ea7f4-0e6f-11e7-9d29-0fe9f8f9b677.png">

<img width="290" alt="screen shot 2017-03-21 at 19 51 45" src="https://cloud.githubusercontent.com/assets/1307818/24165090/d9043a18-0e6f-11e7-8ce4-d20bc6ab9c2b.png">

PR #569, opened 2017-11-07
fix mobile header

lislis:
part of #519 

screenshot-2017-11-7 rorganize it find a rails girls study group 2

Logo, main-nav and profile-nav are now visible and usable on every screen size


app/views/layouts/application.html.erb
```diff
@@ -5,14 +5,13 @@
   <title>RoRganize.it | Find a Rails Girls Study Group</title>
   <%=  stylesheet_link_tag (params[:redesign] ?  'redesign' :  'application'), media: "all" %>
   <%= csrf_meta_tags %>
-  <link href='//fonts.googleapis.com/css?family=Pacifico' rel='stylesheet' type='text/css'>
```
sareg0: Nice!

app/views/layouts/application.html.erb
```diff
@@ -5,14 +5,13 @@
   <title>RoRganize.it | Find a Rails Girls Study Group</title>
   <%=  stylesheet_link_tag (params[:redesign] ?  'redesign' :  'application'), media: "all" %>
```
sareg0: Do you think we can also get rid of this? We can start a new issue for it, so as not to bloat this one.

sareg0:
Thanks soooo much for this. Since there is a fair bit of re-organization of the files, could you point out the fixes that solve the particular UI issues?

app/views/layouts/application.html.erb
lislis: Ah, good catch! I can definitely remove that

app/assets/stylesheets/custom/_nav.scss
```diff
@@ -0,0 +1,154 @@
+// Header + Navigation
+
+$navbar-height-reduced: 78px; // some weirdness on smaller screens
+
+nav.navbar {
```
lislis: @sareg0 most things happen here. I changed the flex-order of the elements that appear in the header to better fit smaller screens. Some elements were also hidden on mobile devices which are now always visible

lislis:
@sareg0 sure thing! I left a comment in the files :)

sareg0:
LGTM

PR #573, opened 2017-11-14
change thumb size for groups index, fixes #466

### LINKED ISSUE #466: Thumbnails on groups#index have the wrong size

lislis:


til:
Thanks for fixing this year-old issue!

lislis:
high five

PR #578, opened 2017-11-14
Use mapbox for geocoder

### LINKED ISSUE #570: Cache geocoder results
I just needed to set this up for a different project that hit the Google geocode API limit.
We don't make super heavy use of maps but maybe it's interesting to some to try it out?
I followed this guide https://github.com/codeforamerica/ohana-api/wiki/Customizing-the-geocoding-configuration but it's very Heroku specific (other project runs on Heroku) but maybe we can get it to run on Uberspace as well.

vsmart:
Closes #570 

todo:
* [x] switch geocoder provider to mapbox
* [x] fix geocoder tests and test fixtures
* [x] add a way to store the api key securely and privately

to deploy: 
* [ ] share mapbox access with folks
* [ ] merge
* [ ] put key on server 
* [ ] git push production

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ pickle-email-*.html
 .project
 .DS_Store
 config/secrets.yml
+.auth
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,3 @@ pickle-email-*.html
 .project
 .DS_Store
 config/secrets.yml
-.auth
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -122,7 +122,7 @@ GEM
     friendly_id (5.2.3)
       activerecord (>= 4.0.0)
     fssm (0.2.10)
-    geocoder (1.4.5)
+    geocoder (1.4.6)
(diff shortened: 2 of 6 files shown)


app/models/group.rb
```diff
@@ -46,7 +46,7 @@ class Group < ActiveRecord::Base
   validates :country, presence: true
 
   mount_uploader :picture, PictureUploader
-  geocoded_by :location
+  geocoded_by :location, :lookup => :mapbox
```
til: I know we are super inconsistent throughout our codebase with this, but let me nevertheless suggest to use the new Ruby hash syntax, at least for newly added code, ie `lookup: :mapbox`

.gitignore
```diff
@@ -19,3 +19,4 @@ pickle-email-*.html
 .project
 .DS_Store
 config/secrets.yml
+.auth
```
vsmart: this should be removed

sareg0:
@vsmart I haven't done the part about sharing mapbox access with the people. Is that something you were planning on doing?

PR #579, opened 2017-11-21
Don't allow http connections during specs

vsmart:
This disables http connections when running specs to ensure tests are run against the fixtures, not the live apis. Our tests shouldn't depend on the reachability of the apis to work. 

Verified that it works by disabling wifi. Tests in master fail with 'could not connect to geocoding api', tests in this branch work.


til:
👏 👏 👏

PR #586, opened 2017-12-12
Tweak some general styles

lislis:
Mostly things that differ from the original design 
screen shot 2017-12-11 at 18 51 41
screen shot 2017-12-11 at 18 51 56


lislis:
screen shot 2017-12-13 at 04 22 23
screen shot 2017-12-13 at 04 22 41
screen shot 2017-12-13 at 04 23 13

app/views/posts/show.html.haml
```diff
@@ -1,18 +1,18 @@
...
+  .action-buttons
+    - if current_person.admin?
+      = link_to 'Edit post', edit_post_path(@post), class: 'btn btn-teal'
+      = link_to 'Delete post', @post, method: :delete, data: {confirm: 'Are you sure?'}, class: 'btn btn-pink'
+
   %header
     - if current_person.admin? && @post.draft?
       %p.post-draft
         It's a draft!
-    %h2.post-title
```
sareg0: I'm not a super huge fan of using the `h2` class for an `h1`... I wonder if there is a way we could make it better. If this is a more wide ranging issue of typographic inconsistencies I'm happy to tackle it in a PR by itself. wdyt?

lislis: True, it's not the nicest class name. The problem is that a lot of the headline styles are applied directly to the h2 element and I was just lazy and made a class name that reflects those styles. 
If you want to tackle the typographic aspects and clean them up a bit please do :D 💃

lislis:
I updated and rebased. I think it's good to go

sareg0:
I am happy with these changes. I am going to merge them

PR #598, opened 2018-01-24
add github handle to profile page

Seunadex:
Add github handle to person profile

Github handle now displays on when user sign in with github

#### Do you have any open Questions you need to get answered
None

Person profile page

#### Screenshot
screen shot 2018-01-31 at 12 50 16 am

screen shot 2018-01-31 at 12 58 15 am


app/models/person.rb
```diff
@@ -64,6 +64,7 @@ def self.from_omniauth(auth)
       person.password = Devise.friendly_token[0,20]
       person.name = auth.info.name   # assuming the user model has a name
       person.remote_picture_url = auth.info.image # assuming the user model has an image
+      person.github_handle = auth.info.nickname
```
lislis: Would this mean that the GitHub handle is automatically set when a user authenticates with GitHub. That would be super convenient

app/views/shared/_github.html.haml
```diff
@@ -0,0 +1 @@
+= link_to object.github_handle, 'http://github.com/'+ object.github_handle, html_option = { target: '_blank' }
```
lislis: Could you make this `https://github.com/` please?

db/migrate/20180124081446_change_username_to_github_handle.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeUsernameToGithubHandle < ActiveRecord::Migration
+  def change
+    rename_column :people, :username, :github_handle
```
lislis: It would be nice to have just one migration for one new database column. Could you re-write this as just one migration and check that in?

db/schema.rb
```diff
@@ -11,7 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170509182902) do
+ActiveRecord::Schema.define(version: 20180124081446) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
```
lislis: There seem to be some extra comments and attributes that appeared in the `schema.rb`. I'm not sure what their all about, especially the `using: btree`. Do you have an idea @til ?

spec/models/person_spec.rb
```diff
@@ -224,6 +225,7 @@
         expect(auth_person.name).to eq 'Buffy Summers'
         expect(auth_person.uid).to eq '123456'
         expect(auth_person.provider).to eq 'github'
+        expect(auth_person.github_handle).to eq 'seunadex'
```
lislis: Nice!

app/models/person.rb
Seunadex: @lislis Yes , the GitHub  handle would be extracted immediately and automatically set when a user logs in with GitHub and is being displayed on his/her profile.

db/migrate/20180124081446_change_username_to_github_handle.rb
Seunadex: Yeah, i wasn't sure the name to give the column initially so i had to change the column name later, but i will refactor that and make it just one migration.

Seunadex:
@lislis All you comments and concerns have been addressed.

lislis:
Awesome, looks perfect!  Thank you!
awesome job

PR #606, opened 2018-02-11
Optimize index action of pages_controller

noordean:
Query Optimization in pages_controller and dashboards_controller

None

No

Index page

### DIFF
--- a/app/views/pages/index.html.haml
+++ b/app/views/pages/index.html.haml
@@ -25,11 +25,11 @@
 .hero
   %h2 Discover and meet
   %p
-    = pluralize(@students_count, 'student')
+    = pluralize(@membership_count.fetch('StudentMembership', 0), 'student')
     and
-    = pluralize(@coaches_count, 'coach')
+    = pluralize(@membership_count.fetch('CoachMembership', 0), 'coach')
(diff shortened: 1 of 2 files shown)


app/views/pages/index.html.haml
```diff
@@ -25,11 +25,11 @@
 .hero
   %h2 Discover and meet
   %p
-    = pluralize(@students_count, 'student')
+    = pluralize(@membership_count['StudentMembership'], 'student')
     and
-    = pluralize(@coaches_count, 'coach')
+    = pluralize(@membership_count['CoachMembership'], 'coach')
```
PragTob: This is a hash, right? I'd prefer we use `fetch` here and above so that we fail hard and early in case the value shouldn't be there :)

noordean: @PragTob I have updated the PR. Kindly check if that's what you meant. Now I know the basic difference between .fetch() and [] on hashes, thanks for that. I later realized that .fetch() can take a second parameter which would be a default value instead of raising an error, hence reason for specifying 0 as the default value.

PragTob:
Thanks! 🎉

PR #608, opened 2018-02-20
updates exception notification gem so it doesn't cause errors in prod

sareg0:
No, it is something we discovered when trying to deploy. Thanks @lislis for catching!

nothing extraordinary, but needs a bit of expediting. Had to update the gem because we were getting the following error about the error_grouping.

```
/gems/exception_notification-4.2.1/lib/exception_notifier.rb:59:in `register_exception_notifier': Invalid notifier 'error_grouping' defined as true (ArgumentError)
```

We fixed the missing comma and update the gem, because that is what the Gem issues said to do. 
https://github.com/smartinez87/exception_notification/issues/384

We tested this by running the server locally with `RAILS_ENV=production`. This allowed us to reproduce the error we saw on the server and to 'verify' that the changes worked

I only updated exception_notification gem, but there are a fair few other change with the gemfile lock. I hope this is ok 🤞 

Check that when the code is deployed the app works, and that the migrations run on the app. You may need someone who can ssh into our server to do this (I can for instance)


lislis:
rihanna yay

PR #609, opened 2018-02-20
move arrow images to public folder, close #594

### LINKED ISSUE #594: Can't find arrow images
After deploying a change we notice that some images were coming up 404

screen shot 2018-01-16 at 21 52 38

Locally the path is `/assets/select-arrow-flipped.png`, but it looks like the image is in the folder `/assets/images/...`

Is this an asset pipeline thing? Either way, we wanted to report it so we can fix it! \o/

lislis:
#594
It seems in the way we organize our .sass/ .scss files is not what Rails/ the asset pipeline expects. For that reason is seems that `image_path` and similar asset helpers don't resolve.
I hope moving the assets to just `/public` will help as a work-around.

n/a

n/a

Look out for the arrow indicators in the select form fields


sareg0:
I tried to help with testing by running the production env in my local rails server, but unfortunately/fortunately we rely on environment variables in production environment that aren't set locally so 🤷‍♀️ 
Seems good to me, I'm happy to try it out

PR #610, opened 2018-02-20
fix tabs on group#show

lislis:
I think I broke group tabs (topics and learning resources) and this PR fixes it.

n/a

n/a

n/a


sareg0:
LGTM

Issue #525, opened 2017-02-26
Show filter options


katrin-k:
On the groups and the people overview pages, there is this city/country filter. 
Once a user sets a filter, this selected option is not shown anymore. 

The minimum should be: the form shows the selected option.
Bonus: The headline "Groups overview" is changed to "Groups overview: Filtered by <selected option>"

lislis:
Started PR #613

lislis:
Also applying this to people overview in #628

lislis:
Both are merged! Closing!

PR #613, opened 2018-02-20
Indicate param filters

lislis:
#525

I started with the filter in groups#index.
People#index still needs to be done

If anyone has opinions on how to 'beautify' this, I'm all ears :D

n/a

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -1,4 +1,3 @@
-
 require 'spec_helper'
 
 describe 'markdown helper', :type => :helper do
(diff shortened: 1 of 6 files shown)


lislis:
You filter groups by city or country or both, it looks like this:

screenshot-2018-2-20 rorganize it find a rails girls study group 1

spec/helpers/application_helper_spec.rb
```diff
@@ -17,3 +16,20 @@
...
+
+describe '#country_name_from_code' do
+  it 'returns a single country name from a single country code' do
+       country = "FR"
+       expect(helper.country_name_from_code(country)).to eq('France')
+  end
+end
+
+describe '#overview_filter' do
+  it 'returns the title for a given overview page, depending on filter params' do
```
sareg0: I'm not sure if a context would be good here, but just something to consider, e.g.

```
context "when only country/only city/both country and city are given as parameters"
  it "returns the string with only the country/the city/the country city in it"  do
    bleep bloopity bloop
  end
```

I guess this might add a fair few lines of code for not really any gains.

spec/helpers/application_helper_spec.rb
```diff
@@ -17,3 +16,20 @@
     expect(helper.country_names_for_select(countries)).to eql [["Germany", "DE"], ["Canada", "CA"], ["United States of America", "US"]]
   end
 end
+
+describe '#country_name_from_code' do
+  it 'returns a single country name from a single country code' do
+       country = "FR"
```
sareg0: looks like one tab too many indented on line 22 and 23

sareg0:
looks really nice! I like the helper too! Also it's awesome to see so many ui tests \o/

LGTM!! I am happy to move this puppy forward 🐶

lislis:
Addressed the comments!

PR #618, opened 2018-03-20
Upgrade rails to 5.0.7

lislis:
We ran into errors in production, since the newest pg version (1.0.0) didn't play well with rails 5.0.5.
Upgrading rails to 5.1.5. fixes it. Tested in production :innocent:


sareg0:
LGTM. I tested this with Lisa so I'm confident we figured it out :)

lislis:
The errors in Travis relate to #619 
Do not merge before we figure this out
clueless

PR #623, opened 2018-04-03
Authenticate before group:new and group:create

### LINKED ISSUE #617: Create group without current_person is broken
The site provides the option of creating a group without being signed in. 

When I attempt to create a group not signed in, I get an error: 

from https://github.com/rubycorns/rorganize.it/blob/master/app/controllers/groups_controller.rb#L28

```

NoMethodError at /groups

undefined method `join!' for nil:NilClass
```

because `current_person` is `nil`. 

So possible solutions are:
1. disable the option to create a group without being signed in? 
2. add a code path that just creates the group, no user associated 
3. forces the user to also create a user-account while creating a group

vsmart:

fixes #617 

- add a test that verifies that without current_person you cannot access these methods.

cc @sareg0

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -2,7 +2,7 @@ class GroupsController < ApplicationController
   respond_to :html
 
   before_action :set_group, only: [:show, :edit, :update, :destroy, :manage_members]
-  before_action :authenticate_person!, only: [:edit, :update, :destroy, :manage_members]
+  before_action :authenticate_person!, only: [:new, :create, :edit, :update, :destroy, :manage_members]
   before_action :ensure_can_destroy, only: [:destroy]
   before_action :ensure_member_powers, only: [:edit, :update]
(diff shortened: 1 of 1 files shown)


vsmart:
Ironically the tests are already adapted to this change, we already test for it working when being authenticated before creating a new group : https://github.com/rubycorns/rorganize.it/blob/master/spec/controllers/groups_controller_spec.rb#L12 .... But should still add another test for the non-authenticated case.

sareg0:
Heya @vsmart it looks like we merged a PR from @Seunadex that has also solved the same issue that this PR solves. Here is the link to it: https://github.com/rubycorns/rorganize.it/pull/626

Do you think we could close your PR? Do you feel like the other PR achieved what you were hoping to do in this one.

vsmart:
hi @sareg0, I just tested in on production and can still reproduce the error. Was #626 deployed already? If no, we should deploy it and test again. If yes, we still need to address this issue.

vsmart:
Closing, this was added in #626!

PR #631, opened 2018-04-18
use flexbox to make footer sticky

lislis:
nope, I noticed while using the live site

n/a

n/a

Pages with only few elements in the main content. Previously the footer was jumping, now it should stay at the end of the page


sareg0:
LGTM

Issue #597, opened 2018-01-17
Rearrange person profile page


lislis:
We cleaned up the group profile page, but the person profile needs some love!

PR #632, opened 2018-04-24
Rearrange person profile

lislis:
#597 

Here is a pic with the new arrangement of all things that can be in the profile
screenshot-2018-4-24 rorganize it find a rails girls study group


sareg0:
I didn't realise that we are using both scss and sass syntax in our styles... Do you think we should transition them all to one syntax? You don't have to do it in this PR, but thought I would ask.

lislis:
party
Imma merge this!

PR #634, opened 2018-05-06
Return 400 when invalid country param is given, close #633

### LINKED ISSUE #633: Return client-side Error when nonsense country/ city is given
```
An ActionView::Template::Error occurred in groups#index:

  undefined method `name' for nil:NilClass
  app/helpers/application_helper.rb:29:in `country_name_from_code'

```
If a nonsense city or country is given in the location filter, this ^ bit returns a 500.
It seems nasty people try to bring down our site by exploiting this, so we should return a 400 status code here.

lislis:
#633

I'm not sure if this is a good way to do it, but it would fix the get param issue that is currently being targeted.

---
400 page
screenshot-2018-5-6 bad request 400


app/helpers/application_helper.rb
```diff
@@ -1,3 +1,6 @@
+class NoSuchCountry < StandardError
+end
+
```
til: I would not have expected this class to be defined exactly at this location. It's true that it is being used by the ApplicationHelper code below, but otherwise it doesn't have a logical connection to the file `app/helpers/application_helper.rb`.

For some cases such as these custom error classes, it's not really clear (to me) where is the best place to define them. but I think if it's a toplevel class, it would be good to put it into its own file in one of the directories that are in the load path, e.g. `app/models/no_such_country.rb`, or `lib/no_such_country.rb`. Or alternatively, maybe it can be put into a namespace (but which one)? E.g. `Group::NoSuchCountry` - what do you think?

app/helpers/application_helper.rb
```diff
@@ -26,6 +29,7 @@ def country_names_for_select(countries)
   end
 
   def country_name_from_code(country)
+    raise NoSuchCountry unless ISO3166::Country[country]
```
til: There is a minor duplication of the `ISO3166::Country[country]` part between this line and the line below - what about putting that into a local variable or a private method?

public/400.html
```diff
@@ -0,0 +1,51 @@
...
+  }
+  </style>
+</head>
+
+<body>
+  <!-- This file lives in public/404.html -->
+  <div class="dialog">
+    <h1>400</h1>
+    <h2>Bad Request</h2>
+    <p>We cannot procede with your request. Please check if it's valid.</p>
```
til: *proceed

public/400.html
```diff
@@ -0,0 +1,51 @@
...
+    <figure>
+      <img src="sloth400.gif" alt="lazy sloth">
```
til: Yeah ;-)

app/helpers/application_helper.rb
lislis: I agree with putting the class into its own file. I've never done custom errors before, so I'm happy for every suggestion.
I don't know about the namespace. We see the param targeted in the group#index but the same is also possible in people#index. Not sure what shared namespace we could make up?

lislis:
Thank you for the review @til I updated the PR

app/helpers/application_helper.rb
til: Good point regarding the namespace - the solution now is the best I could think of.

lislis:
Selfmerge 
self high 5

til:
:five: !

sareg0:
Thank you soooo much y'all This is really great.

Issue #640, opened 2018-05-24
The GDPR ticket 


lislis:
speakerinnen use this generator https://datenschutz-generator.de/
I guess it's ok to paste German?

lislis:
examples
https://hacken.in/privacy_settings
https://signdict.org/privacy
http://speakerinnen.org/en/impressum
https://diversitytickets.org/privacy

PR #643, opened 2018-06-05
[WIP] gdpr

lislis:
#640 

It's not done! All the content is missing.

What goes in there? What info do we need?


sareg0:
So, I have asked around a bit and we should have a privacy policy. i Think we mention it in there.

Issue #645, opened 2018-06-18
500 error when trying to access nonexistent blog page


lislis:
```[RORGANIZE ERROR](2 times) posts#index (RangeError) "invalid page: -4097"```

There is another (automated) attempt to hack our site via this

PragTob:
:scream: 

you sure it's hacking and not ust crawling?

lislis:
hmm, yah. The exact get request url was
``` https://rorganize.it/posts?page=-4097%20UNION%20ALL%20SELECT%20CONCAT%280x7178707671%2C0x5455657865556e4957557a6e4141454a6f4e6b765558515359556d465872726d447442495645694e%2C0x71706b7671%29%23```

PragTob:
oh yeah that's definitely trying SQL injection. Good catch.

sareg0:
whee!! I wanna do eeet! #security

sareg0:
We have merged a PR to solve this, so closing the issue up. Thanks for reporting @lislis

PR #649, opened 2018-07-10
[645] add 500 error when trying to access nonexistent blog page

sareg0:
Yes, https://github.com/rubycorns/rorganize.it/issues/645

not that i can think of, but let me know

Is this a good way to do solve this problem?

You could run the query in the issue (with the SQL injection attempt) to see if you get a bad request error (instead of us getting an exception email)

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -10,6 +10,8 @@ class PostsController < ApplicationController
     @published_posts = Post.published_descending_order
     page = (params[:page] || 1).to_i
     @published_posts = @published_posts.paginate(page: page, per_page: 5)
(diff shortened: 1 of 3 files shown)


app/controllers/posts_controller.rb
```diff
@@ -10,6 +10,8 @@ def index
     @published_posts = Post.published_descending_order
     page = (params[:page] || 1).to_i
     @published_posts = @published_posts.paginate(page: page, per_page: 5)
+  rescue RangeError
+    bad_request
```
PragTob: I think 404 not found might be a better answer :)

sareg0:
awesome thanks @PragTob!!

PR #653, opened 2018-08-14
adds helper to trim whitespace from params in key controllers

### LINKED ISSUE #629: Phantom link in alphabetical people overview
There is a link in the alphabetical list that links to .. nothing! (`#`)
Should be investigated

This is how the phantom link looks like when hovered

screenshot-2018-4-17 rorganize it find a rails girls study group

sareg0:
Yes, closes #629 

Not too much

We would like feedback on if this is the correct way to do things

Creating and updating a person
Creating and updating a group


app/controllers/people_controller.rb
```diff
@@ -22,6 +23,10 @@ def show
   def edit
   end
 
+  def create
```
vsmart: this method can be removed ♻️

app/controllers/groups_controller.rb
```diff
@@ -113,4 +114,8 @@ def group_params
       :searching_students,
       :searching_location)
   end
+
+  def trim_whitespace
```
vsmart: maybe rename to `trim_params`

app/controllers/application_controller.rb
```diff
@@ -51,5 +51,8 @@ def bad_request
     render file: Rails.root.join('public/400.html'), status: 400
   end
 
+  def trim_params(params)
+    params.each_value {|val| val.strip!}
```
sareg0: The test failure is happening because the file upload can't be trimmed... I thought maybe using the respond_to method from Ruby might help

https://ruby-doc.org/core-2.5.1/Object.html#method-i-respond_to-3F

We could check if it responds to the method ''strip" and trim if it does

sareg0:
@PragTob could we trouble you for a review. Also, anyone else who wants to take a look 💯 ❤️

lislis:
I have a question, if you define `trim_params` in the application controller, shouldn't it be available in the other controllers as well because that's the base class?

sareg0:
Good question, we had a bit of trouble using it in the class that inherits the ApplicationController. It didn't seem to happen automatically. We weren't sure what we were doing wrong at the time. So we went with calling super since we say some examples online. But, this didn't seem correct, so if we should try again we totally can.

lislis:
ah, I think it's because RegistrationController doesn't inherit from ApplicationController.
Maybe putting it in a module would make sense?

sareg0:
ohh... okay, good point. Will try this out.

app/models/concerns/params_helper.rb
```diff
@@ -0,0 +1,7 @@
+module ParamsHelper
+  extend ActiveSupport::Concern
```
PragTob: I think we don't need this

Issue #642, opened 2018-05-29
Make all style files use one format: SASS or SCSS


sareg0:
@lislis made a PR to update some SASS-y SCSS-y files, and we realized we have a mix of the two syntaxes. It would be nice to have one syntax. It's a bit simpler to remember how to write things.

So, please, a show of hands if you have a preference or if you think there is one that is better to choose based on science 👨‍🔬 👩‍🔬 🔬 🔍

sareg0:
I vote for SCSS cause it's the most familiar to people who have been working only with CSS

sareg0:
whee! I think we did those, so I will close it

PR #655, opened 2018-08-21
changes all sass files to be scss

sareg0:
#642 

nothing really, just that it's neat and looks good

nah.

test the person page, a random selection of places that uses the util classes and the maps styles. woohoo!!


lislis:
shipping it!
uh yeah

Issue #667, opened 2018-12-03
My groups tabs should only show when user is logged in


lislis:
I just deployed the `My Groups` feature and noticed the tab also shows up when I'm not logged in.
Clicking it results in a 503, which is correct but not so nice to look at.

We should only show the tab when the user is logged in.

lislis:
@Seunadex since you worked on the original feature, would you like to work on this as well?

Seunadex:
@lislis Yes i will

Seunadex:
it's ready [668](https://github.com/rubycorns/rorganize.it/pull/668)  @lislis

lislis:
Thank you very much! It's online already!

PR #668, opened 2018-12-06
Show my groups tab only when authenticated

Seunadex:
[667](https://github.com/rubycorns/rorganize.it/issues/667)

The group's tab


lislis:
Perfect! Thank you!

Issue #647, opened 2018-07-10
Flash message element 'hides' the others underneath


sareg0:
It looks like the element we fill with the flash message is getting in the way of us clicking navigation elements (after an flash has been rendered). Below is a screenshot outlining where the element sits and what navigation items it obscures after updating an address:

<img width="1000" alt="screen shot 2018-07-10 at 19 45 20" src="https://user-images.githubusercontent.com/1307818/42528404-9e481c18-847b-11e8-85ec-5cbb31b9acd4.png">

It's invisible so you can't see it, but the DOM can ;) 
Let me know if you want more details

Seunadex:
@sareg0 This is a [fix](https://github.com/rubycorns/rorganize.it/pull/669)

lislis:
Done in #669 
Thank you @Seunadex

PR #669, opened 2018-12-10
Remove flash message from DOM

Seunadex:
[647](https://github.com/rubycorns/rorganize.it/issues/647)

Nil


lislis:
Awesome thank you!

Issue #650, opened 2018-07-14
[feature] Group admin should have rights to delete their group


vsmart:
## To reproduce

* Created a new group
* You'll be group admin
* Could not delete the group (or I couldn't find the button?)

Seunadex:
@vsmart  [PR-670](https://github.com/rubycorns/rorganize.it/pull/670)

lislis:
Thanks @Seunadex  for building this!
Done and closing!

PR #670, opened 2018-12-10
Group admin can delete their group

Seunadex:
[650](https://github.com/rubycorns/rorganize.it/issues/650)
delete

Visit a group page as a group admin


app/views/groups/_delete.html.haml
```diff
@@ -0,0 +1,3 @@
+- if current_person && current_person.member_of?(@group)
```
lislis: I think this should be `current_person.admin_member_of?`?

lislis:
yaaaay

PR #672, opened 2019-01-08
Upgrade to rails 5.2. [wip]

vsmart:
#### Is this solving an issue already in our backlog? If so, please mention which one

closes #646 

#### Please outline the areas that should be tested, if there are any:

* [ ] Test locally that the app runs 
* [ ] Tests should pass

Issue #646, opened 2018-07-10
upgrade to rails 5.2


lislis:

sareg0:
I am volunteering myself #productivicorn

lislis:
coming in #680

lislis:
needs to be fixed, too #619

sareg0:
@lislis does #619 need to be solved before your PR can be ready?

lislis:
Yes indeed!

vsmart:
We're running on rails 5.2.3 now.

PR #673, opened 2019-02-19
Upgrade to Rails 5.2

vsmart:

#646 

#672


Gemfile
```diff
@@ -14,7 +14,7 @@ gem "pry"
 gem "rubocop"
 
 group :development do
-  gem 'sqlite3'
+  gem 'sqlite3', '1.3.13'
```
vsmart: This should fix the issue i was seeing - "Gem::LoadError: Specified 'sqlite3' for database adapter, but the gem is not loaded" see  https://github.com/rails/rails/issues/35161 for reference 
cc @lislis

lislis: good point, but can you change it to 
```gem 'sqlite3', '~> 1.3.13'``` ?

Comment (vsmart) [reply Gemfile]: done in https://github.com/rubycorns/rorganize.it/pull/673/commits/71fda07c89aaec7d6ca598db0fe7c9b97c9a939b

File: config/environments/production.rb [hunk 2]
```diff
@@ -14,9 +14,14 @@
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
 
+  # Attempt to read encrypted secrets from `config/secrets.yml.enc`.
+  # Requires an encryption key in `ENV["RAILS_MASTER_KEY"]` or
+  # `config/secrets.yml.key`.
+  config.read_encrypted_secrets = true
```
vsmart: change to false, we're not using this yet

config/environments/production.rb
```diff
@@ -61,24 +66,8 @@
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
   # config.action_mailer.raise_delivery_errors = false
 
-  config.action_mailer.default_url_options = {
```
lislis: I think we still need this to send emails in production

config/puma.rb
```diff
@@ -0,0 +1,56 @@
+# Puma can serve each request in a thread from an internal thread pool.
```
lislis: We use Thin instead of Puma, we can delete this file

lislis:
clap

PR #674, opened 2019-03-05
Update rails from 5.1.0 to 5.1.6 to update all the gems

vsmart:

No issue - but will fix Travis build currently failing on master.


lislis:
excite!

PR #680, opened 2019-08-20
rails 5.2.3

lislis:
#646 

I don't think so?

nope

I ran it locally and it did not complain!


vsmart:
slothlove

PR #710, opened 2021-11-27
[wip] start upgrading ruby version and bump gems

lislis:


lislis:
I just added the app/assets/config/manifest.js because it complained it's missing, but I think I also need to add some asset paths

